PR #2, opened 2015-07-31
Midilearn functionnality

Musinux:
Hi everyone,

Could you consider adding the midi learning functionnality to the master branch?
The vast majority of the user interface is midi controllable.

The work has been done in a manner that you can completely control each parameter of any part of the synth (which has been modified adapted, of course) with the same functions:

void changepar( int parameterNum, double parameterValue);
double getpar(int parameterNum);

You can, by doing so, avoid completely the UI and add another system between.

To midi control some part, shift+right click on it and move the knob to assign to. It's done.
Every parameter that is midi controllable can be stored and restored.

Work has been done for the LFO to be real time controllable, so that each note currently played will be modified on-the-fly if you move a knob or your controller associated.

I will be working on a fuller implementation this next few days.

Have a nice day,
Louis Cherel


Musinux:
I saw some bugs that still need to be patched before being clean.

PR #3, opened 2015-11-01
Yoshimi NSM patch port *ONLY PARTIALLY WORKING*

harryhaaren:
Thanks goes to Jonathan M. Liles for the original patch,
without which this would not have been acomplished.

There are a few // NSM_TODO comments left around the source,
those areas *definatly* need attention, and generally the
whole patch could do with a review.

The code in main.cpp is tempermental - the order in which
the SynthEngine and NSM clients are created is very important
to NSM, and generally not so for Yoshimi (i think..) so its
probably not an ideal solution. Anyway :)

Toodles, -Harry van Haaren

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -152,6 +152,10 @@ else (Boost_FOUND)
     message (FATAL_ERROR "boost headers required, but not found")
 endif (Boost_FOUND)
 
+# Liblo
+pkg_search_module(LIBLO liblo>=0.0.0)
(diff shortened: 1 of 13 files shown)


abrolag:
Hi Harry,

Just spotted this before going to bed. I'll have a look at it tomorrow evening
(I still work sometimes!)

Will.

-- 
Will J Godfrey
http://www.musically.me.uk

abrolag:
Pulled into new branch nsm

PR #9, opened 2016-09-24
Splash

jplloyd:
This is basically a refactoring of the way the splash screen image is handled - still statically included at compile time, but now easily edited by the addition of a source svg file and an export script.

The script uses Inkscape to render a png, and then dumps it, as hex-encoded array data, to a separate file which is #included to the splash data definition. Dimensions are read from the svg and updated in the definition of the corresponding constants.

Note that the script is not be part of the build process (for obvious dependency reasons), but should be run separately when making changes to the svg.

Not the best solution, but a significant improvement when it comes to editability.

This branch has not modified the functionality of the actual splash screen.


abrolag:
On Sat, 24 Sep 2016 15:01:56 -0700
Jesper Lloyd <notifications@github.com> wrote:

> This is basically a refactoring of the way the splash screen image is handled - still statically included at compile time, but now easily edited by the addition of a source svg file and an export script.

OK, got that (by a rather roundabout method!).

One problem is that the script is specific to bash, and on debian the prefered
one is dash - that confused me for a while.

I wonder if a python script might be more reliable?

In order to not spring surprises on anyone I've reverted the *actual*
splash screen to the original (I hacked a copy of the code to work directly from
the png).

I'm not sure how best to present this. I'm reluctant to force a general change
on everyone - especially as there was originaly a vote from a selection of four.

Initially, at least I'm inclined to think of a way to make the generation
process slightly easier for self-builders and see how we go from there.

Thanks for your work. It's nicely done.

I've added just your first name to the changelog.

Will.

-- 
Will J Godfrey
http://www.musically.me.uk

jplloyd:
Fair point regarding bash/dash, I only went for bash because of the existing build push script (which obviously, in hindsight, was not part of the distribution).

For a bash/dash-compatible export, all that is required is to set the shebang target to /bin/sh and make two changes to the script:

Line 10:
-function depCheck {
+depCheck() {

Line 23:
-if [[ "$MISSING_DEP" = true ]]; then
+if [ "$MISSING_DEP" = true ]; then

Since there is no need for any advanced structures or complex logic, I do think this is a task better suited for a shell script (unless there is call for cross-platform compatibility). 

Just for public documentation, I will note that self-builders will never need to use the script unless they specifically set out to alter the splash graphic.

As for the splash image itself, I would suggest simply replace the contents of the svg with an embedding of the old png, and change the page dimensions accordingly (for the sake of consistency). I'm not particularly invested in the image I pushed and mainly used it to test the script. However, if there is a call for a new splash graphic at some point, I will certainly be a contender.

/Jesper

abrolag:
Hi Jesper

I had intended to push out V 1.4.2 with just parts of midi-learn hidden from
view and concentrating on the other features, then sometime next year trying to
push ahead with midi-learn and at the same time querying the list about a new
splash screen for V 1.5.0
Well that was the plan, however two things scuppered that idea!

The first was Jörn coming on board with some excellent work on tooltips, and
the second was that midi-learn has gone far, far, better than I expected (makes
a nice change), so now V 1.5.0 will be a Christmas gift :)

Under the circumstance it would be quite unfair to expect anyone to respond at
such short notice, so what I'm thinking of is to ask sometime in January, with
the target to have a new screen (if that's what people want) in place before
LAC 217. I'm hoping to do another workshop, but even if it's not accepted I'll
be there collaring anyone who stands still long enough :)

TTFN

Will.

-- 
Will J Godfrey
http://www.musically.me.uk

jplloyd:
Hello, and sorry for the late reply

My understanding is that the LAC 2017 will be held in May, so there should
be plenty of time. Given some parameters regarding the nature of the
desired screen, I can start sketching up alternatives right away.

Looking forward to try out the midi-learn functionality and check out the
new tooltips!

Regards

2016-12-11 23:08 GMT+01:00 Will J Godfrey <notifications@github.com>:

PR #10, opened 2016-11-25
Add checkbutton to double range of filter frequency tracking

heuchi:
This allows using values beyond 100% and the exact value of +100% which was not possible.

The PR adds a "x2" checkbox. Will's suggestion on the mailing list was "x1.5", but to me it seemed like  there's hardly enough space in the UI to name it "x1.5" and I thought, there are other synths that go up to 200%, so why not? However, if you don't like this range or anything else in this PR, I can change it.

Anyway, I'd be happy to hear comments.

yoshimidft

### DIFF
--- a/src/Effects/DynamicFilter.cpp
+++ b/src/Effects/DynamicFilter.cpp
@@ -27,7 +27,7 @@
 #include "Effects/DynamicFilter.h"
 
 DynamicFilter::DynamicFilter(bool insertion_, float *efxoutl_, float *efxoutr_, SynthEngine *_synth) :
-    Effect(insertion_, efxoutl_, efxoutr_, new FilterParams(0, 64, 64, _synth), 0),
+    Effect(insertion_, efxoutl_, efxoutr_, new FilterParams(0, 64, 64, 0, _synth), 0),
     lfo(_synth),
     Pvolume(110),
     Pdepth(0),
--- a/src/Interface/InterChange.cpp
+++ b/src/Interface/InterChange.cpp
@@ -4541,6 +4541,12 @@ void InterChange::filterReadWrite(CommandBlock *getData, FilterParams *pars, uns
             else
                 val = pars->Ptype;;
             break;
+        case 10:
+            if (write)
+                pars->Pdoublefreqtrack = (val != 0);
+            else
+                val = pars->Pdoublefreqtrack;
+            break;
 
         case 16:
             if (write)
--- a/src/Params/ADnoteParameters.cpp
+++ b/src/Params/ADnoteParameters.cpp
@@ -50,7 +50,7 @@ ADnoteParameters::ADnoteParameters(FFTwrapper *fft_, SynthEngine *_synth) :
     GlobalPar.AmpEnvelope->ADSRinit_dB(0, 40, 127, 25);
     GlobalPar.AmpLfo = new LFOParams(80, 0, 64, 0, 0, 0, 0, 1, synth);
 
-    GlobalPar.GlobalFilter = new FilterParams(2, 94, 40, synth);
+    GlobalPar.GlobalFilter = new FilterParams(2, 94, 40, 0, synth);
     GlobalPar.FilterEnvelope = new EnvelopeParams(0, 1, synth);
     GlobalPar.FilterEnvelope->ADSRinit_filter(64, 40, 64, 70, 60, 64);
     GlobalPar.FilterLfo = new LFOParams(80, 0, 64, 0, 0, 0, 0, 2, synth);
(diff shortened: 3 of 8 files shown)


abrolag:
For some reason this was bounced to me.

On Fri, 25 Nov 2016 06:42:31 -0800
Jörn Eichler <notifications@github.com> wrote:

> This allows using values beyond 100% and the exact value of +100% which was not possible.
> 
> The PR adds a "x2" checkbox. Will's suggestion on the mailing list was "x1.5", but to me it seemed like  there's hardly enough space in the UI to name it "x1.5" and I thought, there are other synths that go up to 200%, so why not? However, if you don't like this range or anything else in this PR, I can change it.
> 
> Anyway, I'd be happy to hear comments.
> 
> yoshimidft
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/Yoshimi/yoshimi/pull/10
> 
> -- Commit Summary --
> 
>   * Add checkbutton to double range of filter frequency tracking
> 
> -- File Changes --
> 
>     M src/Effects/DynamicFilter.cpp (2)
>     M src/Interface/InterChange.cpp (6)
>     M src/Params/ADnoteParameters.cpp (4)
>     M src/Params/FilterParams.cpp (14)
>     M src/Params/FilterParams.h (4)
>     M src/Params/PADnoteParameters.cpp (2)
>     M src/Params/SUBnoteParameters.cpp (2)
>     M src/UI/FilterUI.fl (11)
> 
> -- Patch Links --
> 
> https://github.com/Yoshimi/yoshimi/pull/10.patch
> https://github.com/Yoshimi/yoshimi/pull/10.diff
> 


-- 
Will J Godfrey
http://www.musically.me.uk

abrolag:
Had time for a quick look at this. Curious. Checking by ear, the tracking changes don't seem to be linear. When set for x2 it seems an octave is about '99' - not that actually means anything! I'll attempt to make some frequency measurements tomorrow at work (if I get time).

Code looks neat and complete, but for future reference our style is for { } to be on their own lines.

It occurs to me that an answer to my concern about resolution could be to turn it into a shift instead of a multiplier. i.e. -64 to +63 becomes 0 to + 127. What do you think?

heuchi:
I don't know github auto pull. Is it the reason you closed this?
Other repos I have contributed to usually close unmerged requests, when they decide that the approach taken or the feature intended is not good. So, it took me a bit by surprise, since we still seem to be discussing this. Do you prefer other ways of discussing code or providing patches?

I went through the math once more and I think these are the important places in the source:
The way the frequency tracking value is converted to a "freqpitch"
<pre>
float FilterParams::getfreqtracking(float notefreq)
{
    return logf(notefreq / 440.0f) * (Pfreqtrack - 64.0f) / (64.0f * LOG_2);
}
</pre>
and how the "freqpitch" is converted to a real frequency:
<pre>
float Filter::getrealfreq(float freqpitch)
{
    if (category == 0 || category == 2)
        return powf(2.0f, freqpitch + 9.96578428f); // log2(1000)=9.95748
    else
[...]
</pre>
This latter can also be used to calculate the "globalfilterpitch" for a given filter center frequency.
The final resulting filter frequency is calculated in ADnote.cpp:
<pre>
    float tmpfilterfreq = globalfilterpitch + ctl->filtercutoff.relfreq
          + NoteGlobalPar.FilterFreqTracking;
    tmpfilterfreq = NoteGlobalPar.GlobalFilterL->getrealfreq(tmpfilterfreq);
</pre>
Assuming that the filtercutoff controller is set to its default (and therefore should add 0) here, I took the two values globalfilterpitch (which is -1.1844245 for 440Hz) to generate graphs for the resulting filter frequency for different freq.track. values:
<table><tr><td>

</td><td>

</td></tr></table>
You can see that this gives a perfect linear relation for the pitch (in cents) starting from A3 for freq.track. = 0 and not reaching completely A5 for 127.
With my change the picture changes as expected:
<table><tr><td>

</td><td>

</td></tr></table>
I hope I didn't mix up anything here, and I know it's probably more important how it sounds. But I just wanted to show you, how I came up with the way I did it and why I'm having problems understanding what you mean with "not linear"

heuchi:
Regarding your idea to shift. That's fine for me as well, since my main point was to be able to set it to exactly 100% and we'd even have a finer resolution beyond 100%. We'd loose the values below -100%, but maybe nobody needs them. So I'd be fine with that.

abrolag:
Sorry, closed in error :(
Reopened now.

I think I'd be happier with that. It's probably more a matter of perception than maths, but then I'm more a musician that a mathematician! Personally I've never wanted negative values. Also once the idea was implemented if it did become an issue for anyone we could have a three position switch.

heuchi:
Never mind ;-)

Ok, I'll change it to the (0% - 200%) setting (probably tomorrow).
Do you have any idea for a good label for that (instead of "x2")?

abrolag:
On Sun, 27 Nov 2016 14:07:07 -0800
Jörn Eichler <notifications@github.com> wrote:

> Never mind ;-)
> 
> Ok, I'll change it to the (0% - 200%) setting (probably tomorrow).
> Do you have any idea for a good label for that (instead of "x2")?

Maybe make it a tiny button with a label that toggles -/+ and 0/+

-- 
Will J Godfrey
http://www.musically.me.uk

heuchi:
I've changed the behaviour (and for now just changed the checkbox label to "0/x").
If this is considered good, I can change the variable/parameter name and UI.

It still sometimes sounds slightly out of tune, but I now have the feeling, that could be caused by the resolution of the filter cutoff freq. selector (it has only 128 steps). So if this freq differs slightly from the actual needed one, the further away from 440Hz the pitch goes, the bigger the detuning gets.
I hope this makes any sense... Sometime saying things in English turns out to be more difficult than expected. ;-)
Anyway, since the filter cutoff freq is certainly a candidate to be shown in Hz/kHz in my other project, we might get an answer to this soon.

abrolag:
I'll check this out this evening. However, I do know that there are odd
psycho-acoustic effects on 'experienced' pitch when covering a wide frequency
range, which is also amplitude dependent!

Also I tried doing some measurements during my lunch break at work. Quite
hopeless unfortunately. Even with the narrowest bad I could set, frequency
was quite unstable and impossible to read :(

P·S.
I appreciate your translating to English. Sadly, I can only manage a smattering
of tourist-grade German and a few words of French and Spanish!

PR #11, opened 2016-12-02
Values with units misc

heuchi:
Warning: Pull request ahead.. ;-)

There are still a few controls missing, but I think it might be a good time to show it to some more testers.
I've done almost all controls which appear in the ADDsynth global and voice windows. There's a graphical representation of the Filter Velocity Sensing Function, which I didn't know how to show numerically.

Looking forward to comments. I hope I found appropriate English terms. If anything is not clear, or looks strange, please tell me.

### DIFF
--- a/src/UI/ADnoteUI.fl
+++ b/src/UI/ADnoteUI.fl
@@ -94,7 +94,7 @@ collect_data(synth, o->value(), (Fl::event_button() | 0xc0), 1, npart, kititem, 
         }
         Fl_Dial voicepanning {
           callback {pars->setVoicePan(nvoice, lrint(o->value()));
-
+o->show_panning_value();
 collect_data(synth, o->value(), (Fl::event_button() | 0xc0), 2, npart, kititem, 0x80 + nvoice);}
           tooltip {Voice panning - zero/leftmost is Random} xywh {215 5 20 20} box ROUND_UP_BOX labelsize 10 align 4 maximum 127 step 1
           code0 {o->value(pars->VoicePar[nvoice].PPanning);}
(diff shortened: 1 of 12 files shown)


abrolag:
Very pleased with this. I see you have a value range for Q. This seems to faintly rig a bell.
I've also written to Paul to try to find out what he had in mind for this, and also for the amplitude LFO level.

heuchi:
Thanks, glad you like it.
Of course, as it always is, I just discovered two minor problems considering filter Q tooltip.
It was saying that Formant or SVFilter would not use Q which is clearly wrong. It didn't change back the type coming from the LP1 or HP1 analog filters (these don't use Q).
Do you want me to open a PR for the fix now, or should I go on collecting fixes, unless they are critical (like crashes...)?

abrolag:
As long as there are no show-stoppers please carry on as you are.

The 4 sends at the bottom of the main window are volume levels. They are in the partui as they are 'per part' - which is not immediately obvious!

A point to bear in mind is that eventually there will be no direct access to any synth controls. It will all go via the ring buffers, but it *should* be possible to get current values from FLTK's valuators as these should always be updated.
It's not an issue yet, as gui read and write aren't fully implemented. I can't instigate read from the gui to the gui until I have made writes actually write!

heuchi:
The three functions currently using synth parameters should be easy to change. The filter should be able to just read from the other widgets. I'll put this on my ToDo list and see if it works.
The only thing we might need, is adding a property to LFOUI about its modulation target. This is needed to show the correct value for the LFO depth and cannot be retrieved from other controls (unless you want it to parse the label text...).

abrolag:
The LFO type is defined in the ui by the group integer. We could probably make
this public. Is that what you meant?
0 = amplitude
1 = frequency
2 = filter

heuchi:
Indeed. It should be easy then. I don't think we need to make it public for the tooltip system. I'll give it a try.

PR #12, opened 2016-12-05
Debugging show values

heuchi:
Fixing some things:
Show effective filter freq for SVFilter
Do not use Params classes, but rather use other UI elements where needed
Correctly initialize filter Q tooltip

Since further work in this area probably interferes with these fixes, I open this PR.


abrolag:
Will look at this later this evening - just committed more midi-learn development (and a bugfix)

abrolag:
OK, this is fine. Merged into master.

PR #13, opened 2016-12-11
Values tooltips effects

heuchi:
This adds tooltip values to Echo, Reverb, Chorus, Phaser, EQ and Distortion effects.
I hope still in time for the rc...


abrolag:
Very pleased with this again, and nicely commented code too.
So that's rc2 now :)

Will.

-- 
Will J Godfrey
http://www.musically.me.uk

PR #15, opened 2017-01-15
Adds French translation of the menu item

trebmuh:
(when used with FR locales)


abrolag:
Thanks for this.
As I was already editing several files in doc & desktop it was simpler to just copy the two lines in.
I'll leave this open for a while so others can see it's been dealt with.

trebmuh:
@abrolag : no worries. Thanks for pushing it to master.

PR #17, opened 2017-02-19
Amplitude envelope tooltip fix + envelope freemode values

jplloyd:
Amended conversion function to show the correct amplitude values (sustain volume) when linear interpolation is enabled.

Added conditional drawing of envelope point values in the freemode editor. Shows formatted values in the top right corner when dragging a point.


abrolag:
Thanks for this. I'm in the middle of the CLI number updates at the moment. I'll pull this in as soon as the dust settles.

PR #18, opened 2017-03-04
Tooltip refactoring

jplloyd:
The commit messages give a decent summary of the changes, but I forgot to mention a couple of things:

The sliders now follow the same mousewheel behaviour as the dials, and both now require the mouse cursor to actually be above the widget when handling mousewheel events. Previously the behaviour was inconsistent for dials depending on their position within a group, and sliders could receive mousewheel events even when the cursor was in a completely different location, and regardless of which widget had the focus last.

The filter velocity sense graphic is now refurbished, drawn at 128x64 with a simple background grid.

The bandwidth scale message is a bit of a placeholder for the moment. Although technically accurate, it doesn't really explain what's going on. I will create a graphical representation to replace it before too long.


abrolag:
Once again, thanks for some very nice work :)

PR #24, opened 2017-03-21
Dynamic tooltip delay behaviour

jplloyd:
Adds delay mechanism to dynamic tooltips reflecting that of regular ones. This fixes the (slightly annoying) flickering of tooltips when moving the cursor across multiple controls.

New behaviour:
When hovering, wait for a set amount of time until the tip is shown, delay time depending on whether another tooltip was shown recently. However, show the (formatted) tooltip immediately when changing the value (i.e, pressing, dragging, using the mousewheel).

The conditional delay mechanism does not work between regular and dynamic tooltips, meaning that a regular tooltip having been shown does not affect the delay time for a dynamic tooltip, and vice versa.

It's possible that this patch should wait until after the release, although it technically doesn't add or remove any features.

### DIFF
--- a/src/UI/DynamicTooltip.cpp
+++ b/src/UI/DynamicTooltip.cpp
@@ -36,6 +36,21 @@
 
 #define MAX_TEXT_WIDTH 280
 
+
+// Whether or not a dynamic tooltip was shown recently
+static bool _recent;
+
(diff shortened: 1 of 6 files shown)


abrolag:
These latest ones are small enough to go straight in - there is still some bug hunting going on, so more rc's

PR #25, opened 2017-03-23
Menu bar fixes, visual and functional

jplloyd:
Details in message/changelog. This approach fixes the first submenu being drawn to the left of the window when selected. To draw the top border as well as bottom, just remove the offsets from the y,h parameters in the draw_box call.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,11 @@
 yoshimi 1.5.1 rc3
 
+2017-3-22 Jesper
+* Master window: Added subclassed Fl_Menu_Bar which only draws the bottom border.
+  Now it is not necessary to use negative coordinates to hide the edges.
+* VUMeter no longer swallows all shortcut/keyboard events.
+* Added missing hotkeys/mnemonics and fixed conflicting ones.
+
 2017-3-21 Will
 * Added correction for startup part connection status from Rob.
 * Set master as 1.5.1 rc3
--- a/src/UI/MasterMiscUI.fl
+++ b/src/UI/MasterMiscUI.fl
@@ -89,6 +89,12 @@ decl {\#include "Misc/SynthEngine.h"} {public local
 decl {\#include "Misc/MiscFuncs.h"} {public local
 }
 
+decl {\#include <FL/Fl_Menu_Bar.H>} {public global
+}
+
+decl {\#include <FL/fl_draw.H>} {private global
+}
+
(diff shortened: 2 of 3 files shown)


abrolag:
Good catch on the shotcut issue - I've been puzzling over that one for ages.

Unfortunately I have to re-instate the negative top menu X position, otherwise 'State' spills over the RH edge. There seems to be no way to remove the LF margin on the first entry, nor to reduce the spacing between menu entries. reducing the font size makes the top line difficult to read.
The only other way to resolve this (and make it look balanced) would be to make the entire window wider, are re position all contents!

jplloyd:
Ah, the font on my system lets the bar render with comfortable margins on both sides. After comparing my own graphics to the available screenshots I can see why it can be an issue (the extra space seems to come mainly from cumulatively loose kerning of the small-case t).

However, if we cannot know which font xft will use on an end users system, surely even the current margin is no guarantee that the rightmost menu item will be drawn within bounds? Though it would seem to guarantee that the leftmost menu will be drawn outside of bounds.

My solution would be to move the items in the State menu to the Yoshimi menu to free up enough space for pretty much all reasonable Helvetica substitutes. It would also reflect the common pattern of having the first menu handle save/open/recent for the working session/file/document etc.

I will at least try this out to see if it looks ok.

abrolag:
On Thu, 23 Mar 2017 11:40:48 -0700
Jesper Lloyd <notifications@github.com> wrote:

> My solution would be to move the items in the State menu to the Yoshimi menu to free up enough space for pretty much all reasonable Helvetica substitutes. It would also reflect the common pattern of having the first menu handle save/open/recent for the working session/file/document etc.
> 
> I will at least try this out to see if it looks ok.

I see your point, but if it was going to be moved, then I think the entire menu
tab should be moved as a submenu. As separate items, it would lose continuity,
and one of the reasons we took all of these out of the Yoshimi menu was because
with such a long list it became very difficult to recognise individual items
and their relationship to each other. 

-- 
Will J Godfrey
http://www.musically.me.uk

abrolag:
On the other hand, maybe it's Scales that should move. All the others are
related, and Scales are seldom changed.


-- 
It wasn't me! (Well actually, it probably was)

... the hard part is not dodging what life throws at you,
but trying to catch the good bits.

jplloyd:
A fair point, I will consider frequency of use and the clutter factor when testing different configurations.

PR #28, opened 2017-04-26
Fix invalid desktop file

akien-mga:
Fixes this error:
```
- invalid-desktopfile /usr/share/applications/yoshimi.desktop file contains multiple keys named "GenericName" in group "Desktop Entry"
```


trebmuh:
Good eyes. My bad.

abrolag:
Good catch. I missed that too :(

PR #33, opened 2017-07-25
Formant filter UI: graphical editing + tooltips [FIXED]

jplloyd:
Semantically identical to #32 w.t. exception of a previously omitted crucial whitespace character, which lead to a silent code generation error (and subsequent compilation mayhem).

Changes are now also tested against libfltk1.3.0


abrolag:
Thanks. All fine now :)

PR #35, opened 2017-08-26
fix crash due to the max memory lock

iurien:
Yoshimi LV2 plugin crashed if there is not enough memory to lock. Since the pointers were not set to NULL in the constructor, sometimes there will be an attempt to free an invalid memory. Also if the synth init fails, there is a still attempt to use the engine.


abrolag:
Good catch. Thanks. Entirely down to me, forgetting to NULL those :(
I'll merge this in but will change the Log call so that the message is not hideable - if there's a fatal error the user needs to know what it is!

PR #37, opened 2017-10-20
Fix build on systems using the musl libc

newbluemoon:
On musl systems there is no `random_r` function so my suggestion is to replace it with the `random` function in those cases. But I’m wondering if `random_r` could be replaced in general by `random`?

Additionally musl also doesn’t have argp, but there is a [argp-standalone](https://www.freshports.org/devel/argp-standalone/) which would have to be added to the build dependencies.

And last `PATH_MAX` is defined in `limits.h` so I added this include to src/Misc/MiscFuncs.cpp.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -118,6 +118,22 @@ else (PKG_CONFIG_FOUND)
     message(FATAL_ERROR "pkg-config required but not found")
 endif (PKG_CONFIG_FOUND)
 
+include (CheckFunctionExists)
+check_function_exists(random_r HAVE_RANDOM_R)
+
+# musl systems don’t have random_r() and also need argp-standalone
+if (HAVE_RANDOM_R)
+    add_definitions(-DHAVE_RANDOM_R=1)
+else(HAVE_RANDOM_R)
+    add_definitions(-DHAVE_RANDOM_R=0)
+    find_library (ARGP_LIB argp REQUIRED)
+    if (ARGP_LIB)
+        message(STATUS "Found argp")
+    else(ARGP_LIB)
+        message(FATAL_ERROR "argp required but not found")
+    endif(ARGP_LIB)
+endif(HAVE_RANDOM_R)
+
 # libz
 set (CMAKE_REQUIRED_LIBRARIES z)
 check_c_source_compiles (
@@ -496,6 +512,10 @@ set(ExternLibraries
 add_executable (yoshimi ${ProgSources} main.cpp)
 
 target_link_libraries (yoshimi ${ExternLibraries})
+
+if (NOT HAVE_RANDOM_R)
+    target_link_libraries (yoshimi ${ARGP_LIB})
+endif(NOT HAVE_RANDOM_R)
 
 install (TARGETS yoshimi RUNTIME DESTINATION bin)
 
--- a/src/Misc/MiscFuncs.cpp
+++ b/src/Misc/MiscFuncs.cpp
@@ -29,6 +29,7 @@
 #include <iostream>
 #include <string.h>
 #include <mutex>
+#include <limits.h>
 
 using namespace std;
 
--- a/src/Misc/SynthEngine.cpp
+++ b/src/Misc/SynthEngine.cpp
@@ -229,11 +229,16 @@ bool SynthEngine::Init(unsigned int audiosrate, int audiobufsize)
     }
 
     memset(random_state, 0, sizeof(random_state));
+#if (HAVE_RANDOM_R)
     memset(&random_buf, 0, sizeof(random_buf));
 
     if (initstate_r(samplerate + buffersize + oscilsize, random_state,
                     sizeof(random_state), &random_buf))
         Runtime.Log("SynthEngine Init failed on general randomness");
+#else
+    if (!initstate(samplerate + buffersize + oscilsize, random_state, sizeof(random_state)))
+        Runtime.Log("SynthEngine Init failed on general randomness");
+#endif
 
     if (oscilsize < (buffersize / 2))
     {
--- a/src/Misc/SynthEngine.h
+++ b/src/Misc/SynthEngine.h
@@ -137,7 +137,7 @@ class SynthEngine : private SynthHelper, MiscFuncs
         void ClearNRPNs(void);
         void resetAll(void);
         float numRandom(void);
-        unsigned int random(void);
+        unsigned int randomSE(void);
         void ShutUp(void);
         void allStop(unsigned int stopType);
         int MasterAudio(float *outl [NUM_MIDI_PARTS + 1], float *outr [NUM_MIDI_PARTS + 1], int to_process = 0);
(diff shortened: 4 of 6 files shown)


abrolag:
Had a quick look at this and pulled into a side branch. it appeared to build, but unfortunately when running with an 'ordinary' glib environment on debian testing it fails immediately.

gdb reports:
Yoshimi 1.5.4.1 is starting
[New Thread 0x7ffff7fb7700 (LWP 15245)]
[New Thread 0x7ffff7f36700 (LWP 15246)]
[New Thread 0x7fffeec99700 (LWP 15247)]

Thread 1 "yoshimi" received signal SIGSEGV, Segmentation fault.
0x000055555567080d in SynthEngine::random (this=0x555555a1f070)
    at /home/will/yoshimi-code/src/Misc/SynthEngine.h:319
319             random_result = random();

It would seem that cmake hasn't correctly set HAVE_RANDOM_R

I don't have time to go though this myself (cmake changes are always a headache) so can't accept the patch at the moment.

Interesting catch with the missing header in MiscFuncs - been absent a long time, but will definitely be in the next commit, so thanks for that.

newbluemoon:
HAVE_RANDOM_R is now correctly set. The glibc build works again, sorry for the inconvenience. :-)

abrolag:
OK, that seems fine now. However I'm a bit puzzled by the way you've set up HAVE_RANDOM_R in CMakeLists Is there some specific reason? If not, I'll change it to using 'add_definitions' as we've done elsewhere.

I always acknowledge contributors in the changelog so how would you like to be identified?

P.S.
For future reference, Yoshimi has always used 4 spaces for indentation.

newbluemoon:
Actually those were my first attempts with cmake. I followed a tutorial, it worked and wasn’t aware of ‘add_definitions’, so please feel free to change that. :-)

However, I found the reason for the segfault: it’s a name collison. Both SynthEngine.h and OscilGen.h define a `unsigned int random(void)` function which collides with stdlib’s `long int random (void)` which I would like to use. It works with `rand()` but I don’t know the reason for using `random_r` in the first place and there are many controversial opinions so I’m hesitant to change that. What I did was rename yoshimi’s functions to randomSE and randomOG respectively and it worked, too. But I guess the easiest way would be to use `rand()` on non-glibc platforms. I’m not sure how to proceed.

It’s not really necessary to acknowledge my contribution, but if you really want to you can use my github name (I know, could be a better one). I just like to figure things out ;-)

And I’ll be more carful with the indentation! :-)

abrolag:
cmake is definitely "here be dragons" country :)

rand() gives a weaker randomness that random(). Yoshimi uses it very heavily so that's important (it's also why oscilgen and synthengine have different ones.
Also random_r() is re-entrant which means different threads won't fetch numbers in sequence but will have their own. Again, Yoshimi is multi-threaded so if musl doesn't support random_r() the quality of randomness will be slightly poorer, particularly when Yoshimi is used as an LV2 plugin - where people may run several instances.

Finally, all of this is a compromise as we want the functions to return as fast as possible - they are in the audio thread.

newbluemoon:
Thanks for the detailed explanation! I did rename the functions as mentioned above, cleaned up indentation and pushed again. The problem now is just that the glibc build works but not the musl one. HAVE_RANDOM_R doesn’t seem to be evaluated though it was set correctly.

newbluemoon:
Ok, I was being stupid. It’s fixed now. I also got rid of that cmake include file learned how `add_definitions` works :-)

abrolag:
Ok, this is in now.

newbluemoon:
Thanks! :)

newbluemoon:
Please don’t make a new release yet with this patch. I encountered a new problem. I’m terribly sorry! The GUI doesn’t start. It stops when the splash screen is about to be displayed (there is only a grey rectangle). Command line works however. The strange thing is that when I make the glibc build use random() instead of randr_r() everything works fine! I’m trying to find out what’s causing this.

abrolag:
That certainly seems very odd. the fact the CLI still works suggests it's only the GUI that's freezing not the whole synth. Do you get sound when sending MIDI?

It works fine here on two different machines (but both running under debian). I've made a couple of small style adjustments but they shouldn't be significant, and the whole lot is now in the current master - I like to keep that fairly stable, but it's not a release so the distros won't pick it up.

newbluemoon:
I do get sound when sending MIDI. Another thing is that when exiting the CLI when jack is still running Yoshimi freezes, but when I close jack before exiting everything is fine.

abrolag:
On Sat, 21 Oct 2017 11:26:50 -0700
newbluemoon <notifications@github.com> wrote:

>I do get sound when sending MIDI. Another thing is that when exiting the CLI when jack is still running Yoshimi freezes, but when I close jack before exiting everything is fine.

This is just totally screwy! It looks like something is trashing the splash
screen which is then locking the GUI (shouldn't be possible). I suppose you
could go into the config file and manually disable splash.

Have you possibly got bits of conflicting libraries installed?

What happens if you run earlier version of Yoshimi?

-- 
Will J Godfrey
http://www.musically.me.uk

newbluemoon:
Disabling the splash didn’t change anything but using version 1.4.1 did: it works! :-) Next I’m going to find the latest working release and then we’ll see what changed :-)

newbluemoon:
The latest working relase is 1.5.0.

abrolag:
On Sat, 21 Oct 2017 22:37:51 +0000 (UTC)
newbluemoon <notifications@github.com> wrote:

>The latest working relase is 1.5.0.
>

What I find strange is that you had it working, and now it doesn't. Something
must have changed in your setup. I have no problem here at all.

There were a lot of changes in the GUI between 1.5.0 and 1.5.1 so the only way
to identify the point where this problem is triggered would be to go through
the commits - something I can't do if I don't see the fault!

-- 
Will J Godfrey
http://www.musically.me.uk

newbluemoon:
My main system is glibc, musl builds are done in a chroot. I tested the musl build on a headless Raspberry Pi. And because the build succeeded and the CLI worked I assumed everything else worked, too. Unfortunately I was wrong. Then I did some rearranging and attached a monitor.

I wrote a script which makes a couple of builds between two points in the commit history where I know it worked and don’t worked. I haven’t checked them all yet.

newbluemoon:
Found it! It’s [this commit](https://github.com/Yoshimi/yoshimi/commit/426c65f9b639b5237ee632d7c89780a9c233b020). After reverting it the GUI starts on musl systems in 1.5.4.1 and up to now everything works fine. Will do further testing.

[fltk doc](http://www.fltk.org/doc-1.3/group__fl__multithread.html#gaf6f46de66cc183bc5b38cd364aca4cf2) says:

> The lock() method blocks the current thread until it can safely access FLTK widgets and data.
>
> Child threads should call this method prior to updating any widgets or accessing data. **The main thread must call lock() to initialize the threading support in FLTK.**

I don’t know what is meant with ‘leaking’ in the comment, but maybe maybe moving `Fl::lock()` from the main thread isn’t such a good idea and musl is less forgiving than glibc.

PR #38, opened 2017-10-22
Fix GUI on musl systems / better check for argp

newbluemoon:
This PR reverts commit 426c65f to prevent a GUI crash on musl systems. It puts `Fl::lock()` back into the main GUI thread as it is stated in [fltk’s documentation](http://www.fltk.org/doc-1.3/group__fl__multithread.html).

Furthermore I detachted the check for argp from the check for random_r() because they’re two separate things. E.g. what if another libc has random_r() but not argp? The argp check is also more elaborate (taken from elfutils’ configure script).

This time the builds for glibc and musl are really working.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -105,7 +105,7 @@ set (BuildOptionsThisCPU
 
 set (BuildOptionsBasic
     "-O3 -msse -msse2 -mfpmath=sse -ffast-math -fomit-frame-pointer"
-    CACHE STRING "basic X86 complier options"
+    CACHE STRING "basic X86 compilier options"
 )
 
 set (BuildOptionsDebug
@@ -119,20 +119,32 @@ else (PKG_CONFIG_FOUND)
 endif (PKG_CONFIG_FOUND)
 
 include (CheckFunctionExists)
+# musl systems don’t have random_r() and also need argp-standalone
 check_function_exists(random_r randomR)
 
-# musl systems don’t have random_r() and also need argp-standalone
 if (randomR)
     add_definitions(-DHAVE_RANDOM_R)
-    message(STATUS "Found random_r")
-else(randomR)
+endif(randomR)
+
+check_c_source_compiles (
+    "#include <argp.h>
+     int main () {
+         int argc=1;
+         char *argv[]={\"test\"};
+         argp_parse(0,argc,argv,0,0,0);
+         return 0;
+     }" LIBC_HAS_ARGP
+)
+
+if (NOT LIBC_HAS_ARGP)
+    message(STATUS "libc does not have argp")
     find_library (ARGP_LIB argp REQUIRED)
     if (ARGP_LIB)
-        message(STATUS "Found argp")
+        message(STATUS "Found libargp")
     else(ARGP_LIB)
-        message(FATAL_ERROR "argp required but not found")
+        message(FATAL_ERROR "libargp required but not found")
     endif(ARGP_LIB)
-endif(randomR)
+endif(NOT LIBC_HAS_ARGP)
 
 # libz
 set (CMAKE_REQUIRED_LIBRARIES z)
@@ -513,9 +525,9 @@ add_executable (yoshimi ${ProgSources} main.cpp)
 
 target_link_libraries (yoshimi ${ExternLibraries})
 
-if (NOT HAVE_RANDOM_R)
+if (ARGP_LIB)
     target_link_libraries (yoshimi ${ARGP_LIB})
-endif(NOT HAVE_RANDOM_R)
+endif(ARGP_LIB)
 
 install (TARGETS yoshimi RUNTIME DESTINATION bin)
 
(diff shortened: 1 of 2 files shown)


abrolag:
OK, yes this seems fine now.
The original reason for moving Fl::lock() was that some early gui messages could be sent (and correctly acted on) before the main window was *actually* displayed. This resulted in some random numbers appearing in the CLI· These calls were later removed/reorganised, and as far as I can see the problem has disappeared.

newbluemoon:
Ok, thanks again for merging and sorry for any inconvenience :-)

abrolag:
On Sun, 22 Oct 2017 17:35:50 +0000 (UTC)
newbluemoon <notifications@github.com> wrote:

>Ok, thanks again for merging and sorry for any inconvenience :-)
>

No Prob. We all make mistakes. I only get *really* fussy on releases, and even
then some bugs slip through. Welcome aboard :)

-- 
Will J Godfrey
http://www.musically.me.uk

dbtx:
Oh, boy... I only just now saw this.
I have two problems: first, I'm now seeing similar garbage ejected onto console where I ran e.g. Qtractor when it first opens a session including yoshimi_lv2.so, and I wonder if it's related... but the second problem explains the first. The FLTK docs you linked say "The main thread must call lock()..." and our mainGuiThread() is not that thread, they really do mean main()... where I moved the call per the [additional docs](https://www.fltk.org/doc-1.3/advanced.html#advanced_multithreading)-- "all FLTK draw() methods are executed in the main() thread." which for the LV2 plugin is not even ours! If there isn't a nice way to do this, it seems to preclude FLTK from being used for plugins... which goes a long way toward explaining why suil doesn't support it and thus jalv doesn't work with us.

dbtx:
Backing up a bit-- it is possible that officially correct thread usage in FLTK is not absolutely mandatory, that you don't get full-blown UB so long as you don't do particular things before _somebody_ has called Fl::lock(). The fact that Yoshimi ran for a year (without that symptom!) suggests that the LV2 side could be similarly modified and at least behave, even if it isn't correct. My point is still that for suil, LV2, etc., it can never be officially correct even if it isn't apparently broken. 
On the side, it's not necessary to 'explain why FLTK is unsupported in suil' because somebody could offer to add that support, if they thought they knew how to work around this threading gotcha... After a bit of searching, it really looks like nobody ever even tried, or if they studied the problem I'm looking at and saw it as a show-stopper- they aren't talking about it. Or my search strategy is lacking.
I guess I'm volunteering to find out what changed in main() to make the old TTY garbage symptom go away, and repeat it in the plugin, since in there it's impossible to do FLTK "right". 
And on the other side, walking away so we can keep doing it "wrong" in main.cpp is ...unpleasant. I would like to know more about the crash this PR treated and whether it was anyone's fault or a real incompatibility with musl or something else that could be fixed in Yoshimi while this PR merely makes it go back into hiding.

abrolag:
On Tue, 29 Jan 2019 17:46:04 -0800
Rob Couto <notifications@github.com> wrote:

>Backing up a bit-- it is possible that officially correct thread usage in FLTK is not absolutely mandatory, that you don't get full-blown UB so long as you don't do particular things before _somebody_ has called Fl::lock(). The fact that Yoshimi ran for a year (without that symptom!) suggests that the LV2 side could be similarly modified and at least behave, even if it isn't correct. My point is still that for suil, LV2, etc., it can never be officially correct even if it isn't apparently broken. 
>On the side, it's not necessary to 'explain why FLTK is unsupported in suil' because somebody could offer to add that support, if they thought they knew how to work around this threading gotcha... After a bit of searching, it really looks like nobody ever even tried, or if they studied the problem I'm looking at and saw it as a show-stopper- they aren't talking about it. Or my search strategy is lacking.
>I guess I'm volunteering to find out what changed in main() to make the old TTY garbage symptom go away, and repeat it in the plugin, since in there it's impossible to do FLTK "right". 
>And on the other side, walking away so we can keep doing it "wrong" in main.cpp is ...unpleasant. I would like to know more about the crash this PR treated and whether it was anyone's fault or a real incompatibility with musl or something else that could be fixed in Yoshimi while this PR merely makes it go back into hiding.
>

Hi Rob,
I'm very pleased (relieved?) you're looking at this. I seem to remember there
were several related issues about the same time, most not specifically related
to musl. One was the splash screen locking up the GUI for some window managers.
Another was for the splash screen to actually stay visible and on top if (say)
a bank window was set to auto open centre screen. Yet another was that it
blocked all other Yoshimi activity while it was showing!

All the code for this is in main() including Fl::lock(), which of course LV2
doesn't see.

I've been very cautious about making any changes in the LV2 code. I still don't
understand it and have found to my cost it is frighteningly easy to screw it
up :(

Also, these days it should be possible to make all Yoshimi's controls
accessible to the host, but every time I read through the docs and examples -
such as they are - I find I rapidly lose the umm, thread!

Dunno if any of that helps.

Will.

-- 
Will J Godfrey
http://www.musically.me.uk

PR #40, opened 2017-12-14
FreeBSD fixes

t6:
Hi,

in the FreeBSD port of Yoshimi (https://www.freshports.org/audio/yoshimi) we
currently have to apply a patch to it to get it to install things into the right location.

The libz check seems to be broken on FreeBSD 10.x as well.

Both problems are trivial to fix and shouldn't really affect other platforms, so I'm
hoping this can go in.

Thank you!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -606,7 +606,7 @@ install (DIRECTORY ../doc/ DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/doc/yoshimi
         WORLD_READ WORLD_EXECUTE
 )
 install (FILES ${CMAKE_CURRENT_SOURCE_DIR}/../yoshimi-user-manual.pdf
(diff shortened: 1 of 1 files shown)


t6:
> zlib fix is already in current master.

I don't see the zlib fix in master here on GitHub. Has it not been pushed yet?

abrolag:
On Thu, 14 Dec 2017 04:48:34 -0800
Tobias Kortkamp <notifications@github.com> wrote:

>> zlib fix is already in current master.  
>
>I don't see the zlib fix in master here on GitHub. Has it not been pushed yet?
>
This is the weirdest thing I've ever seen !!!

It was shown as added in build number 1205, commit 640e81cfdbb70ff4a41347aa4bb277665acd359a
but since then seems to have disapeared, yet I can see no reference to it being
removed!

-- 
Will J Godfrey
http://www.musically.me.uk

abrolag:
OK it's in.
Thanks.

t6:
Thank you!

PR #46, opened 2019-01-11
Keep build system CXXFLAGS when not selecting a specialized build

schnitzeltony:
I think specialized builds don't need this but in the first place I cannot test them


schnitzeltony:
Seems fixed - so this one is obsolete - thanks

abrolag:
On Sun, 13 Jan 2019 13:05:30 -0800
Andreas Müller <notifications@github.com> wrote:

>Seems fixed - so this one is obsolete - thanks
>
We're always pleased to see people helping out, so appreciate your work even
though we were already working on this.

-- 
Will J Godfrey
http://www.musically.me.uk

PR #50, opened 2019-05-08
Fix build on FreeBSD with GCC

t6:
Include `sys/types.h` for uint definition

On FreeBSD/powerpc64 yoshimi fails to build with GCC otherwise:
```
/usr/local/bin/g++8  -DDEFAULT_AUDIO=jack_audio -DDEFAULT_MIDI=jack_midi -DGUI_FLTK -DJACK_LATENCY -DJACK_SESSION -DMIN_CONFIG_MAJOR=1 -DMIN_CONFIG_MINOR=5 -DYOSHIMI=\"yoshimi\" -DYOSHI_FIFO_DIR=\"\" -I. -I/wrkdirs/usr/ports/audio/yoshimi/work/yoshimi-1.5.10.2/src -I/usr/local/include/freetype2 -I/usr/local/include/cairo -I/usr/local/include/glib-2.0 -I/usr/local/lib/glib-2.0/include -I/usr/local/include/pixman-1 -I/usr/local/include/libdrm -I/usr/local/include/libpng16 -ffast-math -fomit-frame-pointer -O2 -pipe  -fstack-protector-strong -Wl,-rpath=/usr/local/lib/gcc8  -Wl,-rpath=/usr/local/lib/gcc8 -O3   -std=gnu++11 -Wall -D'YOSHIMI_VERSION="1.5.10.2"' -D'BASE_INSTALL_DIR="/usr/local"' -L/usr/local/lib -ljack -pthread -std=gnu++11 -MD -MT CMakeFiles/yoshimi.dir/Params/PADnoteParameters.cpp.o -MF CMakeFiles/yoshimi.dir/Params/PADnoteParameters.cpp.o.d -o CMakeFiles/yoshimi.dir/Params/PADnoteParameters.cpp.o -c /wrkdirs/usr/ports/audio/yoshimi/work/yoshimi-1.5.10.2/src/Params/PADnoteParameters.cpp
In file included from /wrkdirs/usr/ports/audio/yoshimi/work/yoshimi-1.5.10.2/src/Synth/OscilGen.h:33,
                 from /wrkdirs/usr/ports/audio/yoshimi/work/yoshimi-1.5.10.2/src/Params/PADnoteParameters.cpp:35:
/wrkdirs/usr/ports/audio/yoshimi/work/yoshimi-1.5.10.2/src/Misc/RandomGen.h:190:34: error: 'uint' has not been declared
         uint32_t rot(uint32_t x, uint k) { return (x << k)|(x >> (32-k)); }
[...]

PR #52, opened 2019-05-27
Fix build of Yoshimi 1.5.11.2 on FreeBSD

t6:
Yoshimi 1.5.11.2 fails to build on FreeBSD:
```
In file included from src/Interface/RingBuffer.cpp:27:
src/Interface/RingBuffer.h:34:22: fatal error: use of undeclared identifier 'uint'; did you mean 'int'?
        std::atomic <uint> readPoint{0};
                     ^~~~
                     int
1 error generated.
```
This is a similar error as #50 just in a different place.

### DIFF
--- a/src/Interface/RingBuffer.h
+++ b/src/Interface/RingBuffer.h
@@ -23,6 +23,7 @@
 #ifndef RINGBUFF_H
 #define RINGBUFF_H
 
+#include <sys/types.h>
 #include <atomic>
 #include <stdlib.h>
 


abrolag:
Thanks for reporting this. I've now done a search for any others and fixed them all and also properly set the version number, which is now 1.5.11.3 {sigh} so I won't need to merge this in.

t6:
Thank you.  :-) Unfortunately there are still some `uint` left and the build still fails.
```diff
--- src/Interface/RingBuffer.cpp.orig	2019-05-27 19:17:56 UTC
+++ src/Interface/RingBuffer.cpp
@@ -30,7 +30,7 @@
  * it is ESSENTIAL that all buffers and
  * data blocks are powers of 2
  */
-ringBuff::ringBuff(uint _bufferSize, uint _blockSize):
+ringBuff::ringBuff(uint32_t _bufferSize, uint32_t _blockSize):
     bufferSize(_bufferSize),
     blockSize(_blockSize)
 {
--- src/Interface/RingBuffer.h.orig	2019-05-27 19:10:34 UTC
+++ src/Interface/RingBuffer.h
@@ -38,7 +38,7 @@ class ringBuff
         char *buffer;
         uint8_t blockSize;
     public:
-        ringBuff(uint _bufferSize, uint _blockSize);
+        ringBuff(uint32_t _bufferSize, uint32_t _blockSize);
         ~ringBuff();
         bool write(char *writeData);
         bool read(char *readData);
```

abrolag:
I don't know how I managed to miss these, they are right under the others :(
Patch applied
version 1.5.11.4
Thanks for your patience.

t6:
Thank you. Works fine now.  :-) FreeBSD package updated.

PR #53, opened 2019-06-06
Buffer-branch: use pull processing instead of changing the buffersize dynamically

Ichthyostega:
The goal of this refactoring was to avoid sonic problems due to changing the buffer size dynamically in the middle of ongoing calculations. Especially this addresses problems observed with LV2 and the plugin host "Carla".

The solution proposed hereby is to distinguish between
 - the internal calculation chunk size of the engine
 - the size of the current output request by the I/O system (Jack, Alsa, LV2)

The internal calculation chunk size and buffer can be set by the user through the settings pane.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,12 @@
 yoshimi 1.5.11 M
 
+2019-6-6 Hermann
+* Buffer-handling: use pull processing for sound output,
+  and let the SynthEngine always operate with a fixed,
+  configured calcualtion buffer size. This avoids sonic
+  changes due to dynamic buffer size changes, as could
+  happen with LV2, esp. with the Carla plugin host.
+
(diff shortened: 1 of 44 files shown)


abrolag:
Hi Hermann,

Very pleased to see this and for it to be on the user list now.
However, doing the merge through github itself - I'm not familiar with that
process - I managed to inadvertently merge it straight into master instead of
my copy of buffer.

I rather think it would be quite a major operation to sort that out, so in view
of the general improvement and stability of this now (to say nothing of the
fact that is where it is intended to end up) I'm inclined to shrug and leave it
in.

Initial tests look good, and I'll give it a thorough workout over the next few
days.

Will.

-- 
Will J Godfrey
http://www.musically.me.uk

PR #63, opened 2019-09-23
Fix various mistakes

Diomendius:
I ran through Clang's list of warnings with `-Wall` (GCC may also have caught some of these, but I didn't check) and fixed a few logic errors it flagged, plus some weird formatting with `=-` and a use of `delete` which was [undefined behaviour](https://en.cppreference.com/w/cpp/language/delete).

Most of these are edge-cases, but I think I've restored the intended behaviour for each.

The case of `delete` versus `delete[]` could be improved by using `std::unique_ptr<char[]>`, which still calls `delete[]` in case of an uncaught exception (although streams won't throw exceptions by default, so we're safe in this case) but requires a C++11 compatible compiler. I didn't find any existing uses of `unique_ptr` in the codebase, but C++11 is hardly recent and modern compilers no longer even require `--std=c++11` to enable its features.

It should be relatively easy to replace all instances of `new` / `delete` with `unique_ptr`, and unless compatibility with C++03 is wanted, I think it would be a good idea.

### DIFF
--- a/src/Interface/InterChange.cpp
+++ b/src/Interface/InterChange.cpp
@@ -1545,7 +1545,7 @@ void InterChange::mutedDecode(unsigned int altData)
     {
         case TOPLEVEL::muted::stopSound:
             putData.data.control = MAIN::control::stopSound;
-            putData.data.type = TOPLEVEL::type::Write || TOPLEVEL::type::Integer;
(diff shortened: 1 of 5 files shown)


abrolag:
Thanks a lot for these. I've been using gcc and *none* of these appear, yet they are obviously questionable and the third one quite wrong :(

PR #64, opened 2019-09-25
Fix various undefined behaviour

Diomendius:
Testing with Clang's AddressSanitizer and UndefinedBehaviorSanitizer revealed some more bugs, which I have fixed here.

Clang's LeakSanitizer also reports some memory leaks, but I'm not sure they aren't just one-time initializations intended to live until program exit, or maybe even false positives.

MemorySanitizer (which detects uninitialized reads and some other bugs) might reveal more uninitialized variables, but it apparently gives false positives unless *every* linked library is also compiled with MemorySanitizer, which isn't too fun.

I haven't tested Yoshimi very rigorously, but this should take care of a good chunk of the remaining UB/memory bugs.

### DIFF
--- a/src/Interface/Data2Text.cpp
+++ b/src/Interface/Data2Text.cpp
@@ -60,9 +60,6 @@ string DataText::resolveAll(SynthEngine *_synth, CommandBlock *getData, bool add
     showValue = true;
     string commandName;
 
-    Part *part;
-    part = synth->part[npart];
-
     // this is unique and placed here to avoid Xruns
     if (npart == TOPLEVEL::section::scales && (control <= SCALES::control::tuning || control >= SCALES::control::retune))
         synth->setAllPartMaps();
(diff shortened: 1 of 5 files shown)


abrolag:
On Wed, 25 Sep 2019 02:39:01 -0700
Iain King-Speir <notifications@github.com> wrote:

>Testing with Clang&#39;s AddressSanitizer and UndefinedBehaviorSanitizer revealed some more bugs, which I have fixed here.
>
>Clang&#39;s LeakSanitizer also reports some memory leaks, but I&#39;m not sure they aren&#39;t just one-time initializations intended to live until program exit, or maybe even false positives.
>
>MemorySanitizer (which detects uninitialized reads and some other bugs) might reveal more uninitialized variables, but it apparently gives false positives unless *every* linked library is also compiled with MemorySanitizer, which isn&#39;t too fun.
>
>I haven&#39;t tested Yoshimi very rigorously, but this should take care of a good chunk of the remaining UB/memory bugs.
>You can view, comment on, or merge this pull request online at:
>

>  * Fix OOB access in Data2Text.cpp and MiscGui.cpp
Unfortunately I've had to revert this one.
It seems fine at first, but in kit mode, when *not* at item 1, it fails to
report the waveform type as text, when reading the value from the CLI. I
confess I can't work out why it fails though!

-- 
Will J Godfrey
http://www.musically.me.uk

Diomendius:
I see what the problem was. I hoisted the `kititem != 0` check without noticing that this line had `||`, not `&&`:
https://github.com/Yoshimi/yoshimi/blob/9dd3f30475e5d94bd7cec05ce3fdb7c7acf07c55/src/Interface/Data2Text.cpp#L104

It looks like your recent refactoring fixed both the original bug and my mistake.

Diomendius:
Actually, that might have been a red herring. If `insert == TOPLEVEL::insert::kitGroup` can normally be true when `kititem == 0` it's still a bug, but a separate one to the issue you saw.

The bigger problem is that because I added an extra layer to the if tree, if execution reaches this if block and the condition is false...

https://github.com/Yoshimi/yoshimi/blob/9dd3f30475e5d94bd7cec05ce3fdb7c7acf07c55/src/Interface/Data2Text.cpp#L115-L123

...instead of continuing to the next conditional, execution jumps all the way down to the end of the if/else if chain:

https://github.com/Yoshimi/yoshimi/blob/9dd3f30475e5d94bd7cec05ce3fdb7c7acf07c55/src/Interface/Data2Text.cpp#L253-L267

I'm pretty sure someone's written this down as one of the pitfalls of having long if/else if chains instead of a Big Switch or if-then-break, if-then-return or (gasp) if-then-goto chains.

But you replaced all that with an if-then-return structure in adc9a87a5730c8e0823407e45b7e36b65fc08d20, so that's one less potential trap.

abrolag:
On Wed, 25 Sep 2019 19:20:16 -0700
Iain King-Speir <notifications@github.com> wrote:

>I see what the problem was. I hoisted the `kititem != 0` check without noticing that this line had `||`, not `&&`:
>https://github.com/Yoshimi/yoshimi/blob/9dd3f30475e5d94bd7cec05ce3fdb7c7acf07c55/src/Interface/Data2Text.cpp#L104
>
>It looks like your recent refactoring fixed both the original bug and my mistake.
>

Much of this code has grown organically on fairly sketchy basic structure
(probably not a good way to do it) so does get rather complicated. Now it's
pretty settled it easier to improve (for some undefined value of 'easier) :)

However, I very much appreciate your corrections.

-- 
Will J Godfrey
http://www.musically.me.uk

abrolag:
On Wed, 25 Sep 2019 23:06:54 -0700
Iain King-Speir <notifications@github.com> wrote:

>Actually, that might have been a red herring. If `insert == TOPLEVEL::insert::kitGroup` can normally be true when `kititem == 0` it's still a bug, but a separate one to the issue you saw.

The same *apparent* command for (user seen) item 1 does two different jobs. If
it's not in kit mode it sets the actual effect type by name. If it is in kit
mode it selects one for that kit item from one of the three available effects on
that part by number. simples :@)

Actually, there was a bug in the CLI code where it wasn't sending for kit items
- now sorted.

There are an astonishing number of these variation throughout Yoshimi that are
quite obscure. As I find them, I try to make them more obvious.

-- 
Will J Godfrey
http://www.musically.me.uk

PR #65, opened 2019-09-25
Clean up CMakeLists.txt

Diomendius:
This removes redundant code from CMakeLists.txt, replaces some manual dependency searching with CMake built-in modules or pkg-config, and stops ALSA and JACK linker args being passed to the compiler, thus stopping Clang from warning about unused arguments (GCC does not care).

Everything should be compatible with CMake 3.0, which is currently declared as the minimum required version.

Even more boilerplate could be removed by using the `IMPORTED_TARGET` argument to `pkg_check_modules`, which was added in CMake 3.6. Unfortunately, the equivalent support in `find_package` was added in a piecemeal fashion; for instance, the ALSA package creates an imported target starting with CMake 3.12 while the FLTK package still does not do so.

Imported targets can be passed to `target_link_libraries` directly, and will automagically add the appropriate linker and compiler flags in the process. It's a pity support is still so inconsistent.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -437,8 +437,6 @@ endif(BuildWithFLTK)
 add_definitions (
     -D'YOSHIMI_VERSION="${YOSHIMI_VERSION}"'
     -D'BASE_INSTALL_DIR="${CMAKE_INSTALL_PREFIX}"'
-    ${ALSA_LDFLAGS}
-    ${JACK_LDFLAGS}
     -DYOSHI_FIFO_DIR="${FifoDirectory}"
 )
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -205,41 +205,8 @@ endif(BuildWithFLTK)
 set(CURSES_NEED_NCURSES TRUE)
 find_package (Curses REQUIRED)
 
-find_path(Readline_ROOT_DIR
-    NAMES include/readline/readline.h
-)
-
-find_path(Readline_INCLUDE_DIR
-    NAMES readline/readline.h
-    HINTS ${Readline_ROOT_DIR}/include
-)
-
-find_library(Readline_LIBRARY
-    NAMES readline
-    HINTS ${Readline_ROOT_DIR}/lib
-)
-
-#find_package(PkgConfig REQUIRED)
-if(Readline_INCLUDE_DIR AND Readline_LIBRARY AND CURSES_LIBRARY)
-  set(READLINE_FOUND TRUE)
-else(Readline_INCLUDE_DIR AND Readline_LIBRARY AND CURSES_LIBRARY)
-  FIND_LIBRARY(Readline_LIBRARY NAMES readline)
-  include(FindPackageHandleStandardArgs)
-  FIND_PACKAGE_HANDLE_STANDARD_ARGS(Readline DEFAULT_MSG Readline_INCLUDE_DIR Readline_LIBRARY )
-  MARK_AS_ADVANCED(Readline_INCLUDE_DIR Readline_LIBRARY)
-endif(Readline_INCLUDE_DIR AND Readline_LIBRARY AND CURSES_LIBRARY)
-
-mark_as_advanced(
-    Readline_ROOT_DIR
-    Readline_INCLUDE_DIR
-    Readline_LIBRARY
-)
-
-if(READLINE_FOUND)
-    message (STATUS "Readline library found" )
-else(READLINE_FOUND)
-    message( FATAL_ERROR "Readline library not found! Please install development components (libreadline-dev)" )
-endif(READLINE_FOUND)
+# readline
+pkg_check_modules(READLINE REQUIRED readline)
 
 # set platform specific compiler flags
 if (BuildFor0ld_X86)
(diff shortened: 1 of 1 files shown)


abrolag:
On Wed, 25 Sep 2019 03:15:46 -0700
Iain King-Speir <notifications@github.com> wrote:

>This removes redundant code from CMakeLists.txt, replaces some manual dependency searching with CMake built-in modules or pkg-config, and stops ALSA and JACK linker args being passed to the compiler, thus stopping Clang from warning about unused arguments (GCC does not care).
>
>Everything should be compatible with CMake 3.0, which is currently declared as the minimum required version.
>
>Even more boilerplate could be removed by using the `IMPORTED_TARGET` argument to `pkg_check_modules`, which was added in CMake 3.6. Unfortunately, the equivalent support in `find_package` was added in a piecemeal fashion; for instance, the ALSA package creates an imported target starting with CMake 3.12 while the FLTK package still does not do so.
>
>Imported targets can be passed to `target_link_libraries` directly, and will automagically add the appropriate linker and compiler flags in the process. It&#39;s a pity support is still so inconsistent.
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/Yoshimi/yoshimi/pull/65
>
>-- Commit Summary --
>
>  * Do not pass -lalsa and -ljack to compiler
>  * Remove dead code/duplicate logs in CMakeLists.txt
>  * Use CMake&#39;s FindZLIB instead of handrolled method
>  * Use pkg-config to find readline
>
>-- File Changes --
>
>    M src/CMakeLists.txt (115)
>
>-- Patch Links --
>
>https://github.com/Yoshimi/yoshimi/pull/65.patch
>https://github.com/Yoshimi/yoshimi/pull/65.diff
>

I've had to revert this one as it breaks on readline. I remember we had a lot
of trouble with that :(
I'll check through to see how much of the rest of it can be used.

-- 
Will J Godfrey
http://www.musically.me.uk

Diomendius:
That's frustrating. You mentioned earlier that you're running Debian Buster, so I did some digging and it looks like it's actually Debian's fault this time, not that that helps us. Debian bug report [#901650](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=901650) requests that `readline.pc` be included in Debian's readline package, and the reply indicates that readline 8.0-3 includes the change.

The changelog for readline upstream says `readline.pc.in` was added in June 2014, which I think would mean that readline 6.3 first includes it.

Either way, clearly readline having pkg-config support can't be assumed for quite some time, especially as OS distributions can't be trusted to actually install the file even when the included version of readline supposedly has it.

Finger-pointing aside, all the other commits shouldn't cause problems:
- Having LDFLAGS passed to the compiler never actually did anything, so this is safe
- `find_package` already does the log/abort logic, and this only depends on CMake itself
- `FindZLIB` exists in CMake 3.0 and we have `cmake_minimum_required (VERSION 3.0)`

I'll fix up my branch and submit a new PR.

abrolag:
On Wed, 25 Sep 2019 18:26:14 -0700
Iain King-Speir <notifications@github.com> wrote:

>That's frustrating. You mentioned earlier that you're running Debian Buster, so I did some digging and it looks like it's actually Debian's fault this time, not that that helps us. Debian bug report [#901650](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=901650) requests that `readline.pc` be included in Debian's readline package, and the reply indicates that readline 8.0-3 includes the change.
>
>The changelog for readline upstream says `readline.pc.in` was added in June 2014, which I think would mean that readline 6.3 first includes it.
>
>Either way, clearly readline having pkg-config support can't be assumed for quite some time, especially as OS distributions can't be trusted to actually install the file even when the included version of readline supposedly has it.
>
>Finger-pointing aside, all the other commits shouldn't cause problems:
>- Having LDFLAGS passed to the compiler never actually did anything, so this is safe
>- `find_package` already does the log/abort logic, and this only depends on CMake itself
>- `FindZLIB` exists in CMake 3.0 and we have `cmake_minimum_required (VERSION 3.0)`
>
>I'll fix up my branch and submit a new PR.
>

In Yoshimi, cmake and their various interactions with distros has been flaky for
as long as I can remember.

This was the second time debian caused a major failure in cmake. the first
time resulted in *us* being blamed for their mistake and Yoshimi pulled from
their distro for a month :(

Fedora has been the most fussy about correctness - fortunately we are on good
terms with one of the maintainers. I did keep a build of it myself, but really
can't get on with the style, so scrapped it.

Arch is the one that can cause headaches by being too new!

Since we started supporting BSD, freeBSD has thrown up some oddities, but
for some time now that's always been due to them have upstream dependency
errors. Quite often Yoshimi will build on (say) 32 bit systems but not 64 bit -
or vice-versa!

-- 
Will J Godfrey
http://www.musically.me.uk

abrolag:
Hi, did you find time to have another look at this?

Diomendius:
I've been preoccupied of late, but I think my local branch has been ready to merge for a while now. I'm about as certain as I can be without testing that the only regression was the readline change, so I'll open a pull request for what I have so far.

There are a few new commits since what I submitted last time, but none of them should change the script's behaviour.

PR #69, opened 2019-10-13
Fix for OscilGen thread issue and one cleanup

kramlie:
This fixes the OscilGen thread-safety issue and cleans up one unneeded optimization. See the individual commits for details.

The patch looks big, but the vast majority is just moving code around to split OscilGen into OscilParameters and OscilGen, and renaming parameters thereafter. There isn't a whole lot of new code here.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -384,6 +384,7 @@ set (Params_sources
     Params/SUBnoteParameters.cpp  Params/PADnoteParameters.cpp
     Params/Controller.cpp  Params/Presets.cpp
     Params/PresetsStore.cpp  Params/UnifiedPresets.cpp
+    Params/OscilParameters.cpp
 )
 
 set (Synth_sources
--- a/src/DSP/FFTwrapper.cpp
+++ b/src/DSP/FFTwrapper.cpp
@@ -69,7 +69,7 @@ void FFTwrapper::deleteFFTFREQS(FFTFREQS *f)
 
 
 // Fast Fourier Transform
-void FFTwrapper::smps2freqs(float *smps, FFTFREQS *freqs)
+void FFTwrapper::smps2freqs(const float *smps, FFTFREQS *freqs)
 {
     memcpy(data1, smps, fftsize * sizeof(float));
     fftwf_execute(planBasic);
@@ -81,7 +81,7 @@ void FFTwrapper::smps2freqs(float *smps, FFTFREQS *freqs)
 
 
 // Inverse Fast Fourier Transform
-void FFTwrapper::freqs2smps(FFTFREQS *freqs, float *smps)
+void FFTwrapper::freqs2smps(const FFTFREQS *freqs, float *smps)
 {
     memcpy(data2, freqs->c, half_fftsize * sizeof(float));
     data2[half_fftsize] = 0.0;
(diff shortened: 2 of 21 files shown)


kramlie:
Don't merge this, there is something wrong when loading parameters. WIP.

abrolag:
OK, I'll wait... Watching this with *great* interest :)

kramlie:
Alright, fixed that issue, and one preexisting bug as well (the group phase randomness). Check it out!

abrolag:
Unfortunately there is a problem :(
I pulled this into a 'fixes' branch just in case...
There are significant missing harmonics on some sounds. Great Organ shows this up most clearly

kramlie:
You have a knack for finding the problematic instruments! It appears to affect User Base Functions specifically, though I'm not sure why. I'm looking into it though! Thanks for finding this.

kramlie:
Fixed!

abrolag:
I have a group of compositions that i know extremely well and use for serious testing various aspects. As far as i can tell, this branch now sounds identical to master.
One of my compositions pushes Yoshimi very hard, and unfortunately I'm seeing a significant increase in DSP load.
I usually test on a middle-of-the-road dual core machine, at 48k 64 frames, and normally this track gives just 1 or 2 Xruns. With this branch they are well into to 30s and very noticable :(

I've not looked at the code in detail, so have no idea what the cause is. Do you think it likely this can be improved?

kramlie:
Can you try to revert just this commit locally, and see if it helps?

```
git revert f2a9311b3315846b9491d79a213b8dc71d0be4d7
```

In that commit I removed an optimization for two reasons:
* I think the conditions are very rarely met, and as such it's a complicated optimization to keep for little benefit. But maybe I'm wrong!
* It is incompatible with live feedback. This will be considerably more difficult to handle, but let's see the results first, and we can take it from there. Maybe this is not the reason after all

abrolag:
This didn't make any noticeable difference. Below is a link to the archived Rosegarden and Yoshimi files if you want to do comparisons yourself.
www.musically.me.uk/themainevent/GhostTrain.tgz

kramlie:
Ok, that's good actually, because I don't know how we could preserve that optimization together with live feedback!

But interesting that the performance has dropped so much, this is definitely not intentional, nor can I think of any potential hot spots. I'll just have to dig into it. Thanks for the files, that could come in handy!

I'll let you know when this little helper has anything to report!

abrolag:
On Wed, 16 Oct 2019 11:03:15 -0700
kramlie <notifications@github.com> wrote:

>I have made several fixes now, including one so-far-unknown crash. This makes the performance of your composition a lot better on my computer, although I think it is more underpowered than yours, since even now it does xrun considerably. But without these latest fixes it barely even *gets started* before xrunning! At least now it has a chance.
>
>I also did some more rigorous checking of when the `OscilGen->prepare()` function is actually called, which is the heart of the problem here. And I have confirmed that it is only called once at the start of the very first note, and that's it. Which is how it should be (and was before)!
>
>Let me know if this improves it for you!
>
Hi Kristian,
Yes, that's dramatically better, just two very brief Xruns, which seem to be
in similar places to 'normal'. I'll do some more testing later, and then if
nothing nasty shows up, I'll merge into master :)

Will.

-- 
Will J Godfrey
http://www.musically.me.uk

kramlie:
Great!

abrolag:
After pretty extensive checks I can find no detectable differences, so now merged into master.

PR #73, opened 2019-10-21
Improvements to CMakeLists.txt

Diomendius:
Various improvements to formatting, removal of redundant code, replacement of manually locating ZLib with a built-in module in CMake, and a couple of minor bug fixes.

This includes the changes from #65 (which were reverted), minus the pkgconfig and readline related regression.


abrolag:
This seems fine now. However in view of the unexpected issues in the past, in the next push from here I'll include CMakeListsOLD.txt

PR #76, opened 2019-12-29
Live parameter feedback for Yoshimi

kramlie:
This is the followup to [the thread on the mailinglist](https://www.freelists.org/post/yoshimi/Instant-control-feedback), and implements instant audio feedback for nearly all parameters in Yoshimi. Those that don't give feedback are:

* PADsynth oscillator change
  * It's way too expensive to regenerate the oscillator on every change. It would be solvable with a queue and delay approach, but this is very different so I won't implement this now
* Most parameters that add, remove, replace or rearrange whole elements in the audio chain. This covers
  * All controls that enable something (enable an engine, enable filter, enable envelope, etc)
  * Changing the number of stages in a filter, or the type of filter (except the filter inside the oscillator window, this one isn't a real filter and gives live feedback)
  * Changing the number of unison voices
  * Changing modulation type
  * Changing the voice or modulation source
  * Changing the number of formants or sequence size in the formant filter

Possibly I've missed some corner cases in the list. The exceptions above were omitted from live feedback primarily because they allocate new structures, and reallocating on the fly is something I would like to avoid, at least in the first iteration.

The code is relatively finished, and I have done quite a bit of testing on it (@abrolag's GhostTrain track was a big help in pushing Yoshimi and finding bugs!), but it's a big piece of code, so any testing from others is appreciated!

There are also some unresolved questions which I'd like feedback on:

1. Is the list of exceptions to live feedback ok for now? The way I see it, we are making it incrementally better
2. There was some talk on the mailing list about keeping an option around to disable live feedback for specific patches. However, from the responses it didn't seem like anyone is relying on the non-live feedback, and although I supported it then, I've found that adding it has several drawbacks:
   * It's a lot of work to add it, because we need to make almost all parameter access conditional
   * For the same reason, it will reduce performance
   * The non-live feedback in the old version was inconsistent, with some parameters being live, and some not (I know about at least LFO parameters being live), and fixing this would be even more effort
* So my vote would be to make it live only for now, as this seems by far the most useful. The only use case I can see for non live-feedback is if someone relies on changing something via MIDI learn, and then have it change only on the *next* note, not the current one. Which is a pretty esoteric use case. But what do you think?

### DIFF
--- a/src/DSP/Filter.cpp
+++ b/src/DSP/Filter.cpp
@@ -28,7 +28,9 @@
 using func::dB2rap;
 
 
-Filter::Filter(FilterParams *pars, SynthEngine *_synth):
+Filter::Filter(FilterParams *pars_, SynthEngine *_synth):
+    pars(pars_),
+    parsUpdate(pars_),
     synth(_synth)
 {
     unsigned char Ftype = pars->Ptype;
@@ -44,17 +46,10 @@ Filter::Filter(FilterParams *pars, SynthEngine *_synth):
 
         case 2:
             filter = new SVFilter(Ftype, 1000.0f, pars->getq(), Fstages, synth);
-            filter->outgain = dB2rap(pars->getgain());
-            if (filter->outgain > 1.0f)
-                filter->outgain = sqrtf(filter->outgain);
             break;
 
         default:
             filter = new AnalogFilter(Ftype, 1000.0f, pars->getq(), Fstages, synth);
-            if (Ftype >= 6 && Ftype <= 8)
-                filter->setgain(pars->getgain());
-            else
-                filter->outgain = dB2rap(pars->getgain());
             break;
     }
 }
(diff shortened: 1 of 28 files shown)


abrolag:
Seriously impressed with this, both at the amount you've achieved, and the timescale!
I'm fully in agreement with these switches remaining as next-note. Apart from the difficulty of changing them, I've actually made use of them combined with MIDI-learn.

Also, seeing/hearing this in action I'm now happy with the idea of the instant ones always being so. 
There just one switch I'm not sure about being instant, and the is the octave ones, but I have no strong feelings so will go with the majority.

kramlie:
Thanks! What I'm particularly happy about, even though it means nothing to the end user, is that the code ended up smaller than the original, by more than 50 lines! The primary reason is a lot of duplication that used to be in the note initialization functions. Good to have that cleaned up and centralized.

Ideally I would like everything to have live feedback, those that aren't are more for technical reasons. So I would vote for the octave switch having live feedback. Maybe one day they will all behave like that.

kramlie:
@abrolag: I fixed the bug you reported where notes would sometimes have an amplitude off the charts!

abrolag:
... and it's in :)

PR #80, opened 2020-01-22
Fix volume knob on individual voices not working after being zero.

kramlie:
When it hits zero, the voice is killed, but this is no longer
desirable after we got live controller feedback. So we lose a small
optimization, but gain the ability to use the volume knob freely with
for example MIDI learn.


abrolag:
OK, this seems fine - it's not exactly a huge change :)

PR #83, opened 2020-02-26
Various bugfixes

kramlie:
See individual commits for details. Most were found during my own playing around with legato, though one bug [is from the mailing list](https://www.freelists.org/post/yoshimi/Little-bug-with-realtime-updates-in-Subsynth).


abrolag:
Thanks a lot for these. Nice work :)

PR #85, opened 2020-03-03
Remove redundant code

Diomendius:
This removes several redundant pieces of code:
- Several checks that a container is not empty before emptying it
- A check that a list is not empty before calling `remove()` on it
- Several calls to `clear()` on members inside their class's constructor or destructor
- A check that a string does not equal `""` before a while loop with an identical condition

Calling `clear()` on an empty container is a no-op. Calling `remove()` on an empty list is a no-op. When an object is constructed, members not present in the member initializer list are default-constructed, which in the case of containers means a valid and empty container. When an object is destroyed, its members' destructors are also called, which in the case of containers clears the container.


abrolag:
Thanks again for these

PR #86, opened 2020-03-18
Fix legato

### LINKED ISSUE #51: Legato/Portamento transitions are imperfect, causing slight popping
Transitions between notes in Legato mode do not reliably happen smoothly. Depending on the type of synths used, the number of legato transitions chained together, and in the case of ADDSynth, oscillator phase randomness, transients are sometimes introduced at the exact point of the next note's NoteOn.

It seems to me that some aspects of note state aren't preserved when transitioning to the next note. Inconsistencies in phase at the end of one note and the start of the next are the most obvious culprit, although more complex instruments may have other problems, too.

legato_1
Looking at the waveform of a test sequence of octave jumps using "Simple Sound" set to legato mode, the first transition looks perfect. Some popping can be heard, but only due to the instant transition between frequencies. ADDSynth consistently does well on the first transition (with some exceptions when oscillator phase randomness is used).

legato_2
The second transition (back down an octave) looks nearly as good. The transition appears to happen when one note is at 0° phase, but the amplitude drops slightly for the next two peaks, so the phase is likely slightly mismatched, probably by chance.

legato_3
The next transition up an octave shows the problem clearly. The next note appears to start at nearly opposite phase, the amplitude drops for the next several peaks and a pop is audible, although masked by the sudden change in pitch.

Yoshimi appears to crossfade (or otherwise smooth) one note into the next in legato mode, but this only masks the artifacts slightly.

The problem is more audible with portamento enabled (as the popping is obvious without a sudden change in pitch), but portamento itself does not appear to be the cause of the problem.

portamento_1
The first transition in the same sequence, but with portamento enabled. Completely seamless.

portamento_2
The second transition has an obvious artifact.

portamento_3
This time, the third transition is nearly seamless, although the bottom peak just before the 1.390 marker is slightly shallower.

ADDSynth without any "phase randomness" settings enabled consistently generates smooth transitions, but only for the first consecutive transition. Yoshimi clearly works some magic to make sure the ends line up, but forgets something after the first transition. As soon as the last note is stopped, the next legato transition will be seamless again.

With the phase randomness slider in the oscillator waveform dialog set to the far right (63, maximum randomness, per harmonic) ADDSynth no longer reliably generates smooth transitions even for the first in a sequence. With the phase randomness slider set to the far left (-64, maximum randomness, shared across all harmonics), ADDSynth still generates a perfect transition for the first in a sequence, every time.

SUBSynth and PADSynth do not reliably generate smooth transitions at all. It looks like only the crossfading is done to mask the transition.

I think this could prove to be a very hairy problem to solve across the board, considering how complex instruments can be, but it should be possible. The next note should inherit every time-variant aspect of the previous note's state, including phase/position in the waveform/wavetable, modulator and LFO position.

In fact, it seems the most sensible to me to reuse the existing note object entirely rather than trying to copy over every relevant part of one note into a new one, but maybe I'm missing something due to not being familiar with the codebase.

As a side note, I hope this is the most appropriate place to report this bug. Between here, the Sourceforge tracker and the mailing list, it isn't obvious what the preferred location for bug reports is.

Diomendius:
Currently, legato is all kinds of broken (see #51). Crossfading is attempted, but doesn't work correctly and actually causes the kind of harsh transients it attempts to prevent. The current method is overengineered, doing much more work than necessary, and it is fragile enough that trying to play a new note before the crossfade is completed (admittedly it only lasts 5ms) is actually prevented entirely.

This fixes all of that. Legato is entirely reimplemented for all three synth engines, and seems to work fine after testing. Legato with or without portamento should result in a smooth transition in all cases, regardless of the synth engines used.

Portamento no longer triggers harmonic re-randomization in ADSynth, which was one source of glitchy sound, but regular legato still does, as crossfading works now and the change in timbre is probably desired in this case. I chose this behaviour based on the assumption that a plain legato transition is still effectively a new note with no attack, while portamento is equivalent to pitch bend, just controlled in a different way. It may be desirable to be able to disable harmonic re-randomization for plain legato notes, or possibly enable a smooth transition from one timbre to another during portamento.

This works by reimplementing legato as simply setting the parameters of the playing note to the new ones in the NoteOn MIDI message, setting it to start fading in from 0 gain, and cloning the note to another, which is then set to fade out at the same rate.

Like the old method, each note played in legato mode while no others are being played spawns an extra note which is silent by default and gets used as one half of the crossfade. Unlike the old method, the silent note does not perform any computation while it is silent, as it resyncs by copying the relevant internal parameters from the audible note when starting a crossfade. Also unlike the old method, the duplicate note which starts silent (kept track of by `posb` in `Misc/Part.cpp`) is only used for fading out, rather than the last note to fade out being used to fade in and vice versa. As the `posb` note is now a perfect clone of the `posa` note right before a crossfade begins, there is no need for the extra complexity. Portamento does not use posb at all as the smooth change in frequency does not create anything that needs a crossfade to smooth over.

One limitation that I have not yet been able to work around is due to the way PADSynth works. PADSynth generates multiple wavetables of a fixed number of samples but with different base frequencies. It does this so that the tables can be read at the appropriate sample rate for the desired "real frequency" of the note without worrying about aliasing when high frequencies are shifted above the Nyquist limit. Without some form of precomputation like this, the table would have to be low-pass filtered at runtime before being resampled. This might actually be a more desirable method to use in some respects, but there are good reasons to do it this way, too.

PADSynth using multiple wavetables in this way is a problem because there is no connection or possible "splice point" between any pair of tables, so the table a note uses can't be changed seamlessly while the note is playing. I did some experimenting by making PADSynth generate all the tables using the same randomized set of phases for each table rather than randomizing separately for each table, but I quickly found (with the help of the export feature so I could look at the tables directly in an audio editor) that the fact the amplitudes were different for each table made the waveform in each table too different from that of every other even with the phases matched, no matter where you decided to splice them after resampling. It seems obvious in hindsight, but it was still interesting to explore.

The best way to solve the PADSynth issue is probably to crossfade when doing portamento on a `PADnote`, but this would need special logic to do right. The existing portamento logic depends on `basefreq`, so the frequency doesn't match between the halves of the crossfade if you just enable crossfading for PADSynth portamento and turn on portamento for the fading-out note too. A proper fix would likely work in `PADnote` itself, and take into account the possibility of moving across multiple tables in one slide. In fact, a proper fix should probably work for pitch bend as well as portamento.

As I have it currently, portamento just keeps the same wavetable, just as when pitch bend is used. When shifting up a lot in pitch (more than half an octave or so, but it depends heavily on the high-frequency content of the oscillator used to generate the tables) aliasing is audible, and when shifting down a lot in pitch the sound is effectively low-pass filtered.

See https://github.com/Yoshimi/yoshimi/issues/51#issuecomment-598765371 for a detailed description of the previous legato implementation and its bugs.

Closes #51

### DIFF
--- a/src/Params/PADnoteParameters.cpp
+++ b/src/Params/PADnoteParameters.cpp
@@ -610,7 +610,7 @@ void PADnoteParameters::applyparameters()
         samplemax = 1;
 
     // prepare a BIG FFT stuff
-    FFTwrapper *fft = new FFTwrapper(samplesize);
+    FFTwrapper fft = FFTwrapper(samplesize);
     FFTFREQS fftfreqs;
     FFTwrapper::newFFTFREQS(&fftfreqs, samplesize / 2);
 
@@ -641,7 +641,7 @@ void PADnoteParameters::applyparameters()
             fftfreqs.c[i] = spectrum[i] * cosf(phase);
             fftfreqs.s[i] = spectrum[i] * sinf(phase);
         }
-        fft->freqs2smps(&fftfreqs, newsample.smp);
+        fft.freqs2smps(&fftfreqs, newsample.smp);
         // that's all; here is the only ifft for the whole sample; no windows are used ;-)
 
         // normalize(rms)
@@ -666,7 +666,6 @@ void PADnoteParameters::applyparameters()
         sample[nsample].basefreq = basefreq * basefreqadjust;
         newsample.smp = NULL;
     }
-    delete fft;
     FFTwrapper::deleteFFTFREQS(&fftfreqs);
 
     // delete the additional samples that might exists and are not useful
--- a/src/Params/PADnoteParameters.cpp
+++ b/src/Params/PADnoteParameters.cpp
@@ -25,6 +25,8 @@
 
 */
 
+#include <memory>
+
 #include "Misc/XMLwrapper.h"
 #include "DSP/FFTwrapper.h"
 #include "Synth/OscilGen.h"
@@ -586,7 +588,9 @@ void PADnoteParameters::applyparameters()
 {
     const int samplesize = (((int)1) << (Pquality.samplesize + 14));
     int spectrumsize = samplesize / 2;
-    float spectrum[spectrumsize];
+    // spectrumsize can be quite large (up to 2MiB) and this is not a hot
+    // function, so allocate this on the heap
+    std::unique_ptr<float[]> spectrum(new float[spectrumsize]);
     int profilesize = 512;
     float profile[profilesize];
 
@@ -623,11 +627,11 @@ void PADnoteParameters::applyparameters()
         float basefreqadjust = powf(2.0f, tmp);
 
         if (Pmode == 0)
-            generatespectrum_bandwidthMode(spectrum, spectrumsize,
+            generatespectrum_bandwidthMode(&spectrum[0], spectrumsize,
                                            basefreq * basefreqadjust, profile,
                                            profilesize, bwadjust);
         else
-            generatespectrum_otherModes(spectrum, spectrumsize,
+            generatespectrum_otherModes(&spectrum[0], spectrumsize,
                                         basefreq * basefreqadjust);
 
         const int extra_samples = 5; // the last samples contains the first
--- a/src/DSP/AnalogFilter.cpp
+++ b/src/DSP/AnalogFilter.cpp
@@ -58,6 +58,37 @@ AnalogFilter::AnalogFilter(unsigned char Ftype, float Ffreq, float Fq, unsigned 
 }
 
 
+AnalogFilter::AnalogFilter(const AnalogFilter &orig) :
+    type(orig.type),
+    stages(orig.stages),
+    freq(orig.freq),
+    q(orig.q),
+    gain(orig.gain),
+    order(orig.order),
+    needsinterpolation(orig.needsinterpolation),
+    firsttime(orig.firsttime),
+    abovenq(orig.abovenq),
+    oldabovenq(orig.oldabovenq),
+    synth(orig.synth)
+{
+    outgain = orig.outgain;
+
+    memcpy(x, orig.x, sizeof(x));
+    memcpy(y, orig.y, sizeof(y));
+    memcpy(oldx, orig.oldx, sizeof(oldx));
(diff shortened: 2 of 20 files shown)


Diomendius:
One further improvement I haven't made yet would be to do away (at least mostly) with `pos`/`posb` entirely. Currently, two notes are allocated on NoteOn in legato mode when all notes are off/released and these notes are reused for legato until all notes are released again. Now that legato doesn't require any computation on silent notes to keep them in sync, the fading out note doesn't actually need to exist at all after it finishes fading out. The logic to keep the silent note alive can be removed, the logic to start a fade-out on the existing `posb` note can simply clone a new note before calling `legatoFadeOut()`, which can be simplified to no longer copy anything from the other note, as it would already be in sync on construction.

Doing this would require either reinstating the "ignore NoteOn during legato fade" behaviour that `legatoFading` was for, or keeping track of existing fading notes so crossfading can work properly with more than two notes at once.

This would eliminate a note allocation on every first note in legato mode, exchanging it for an extra allocation on every subsequent note until all notes are released again. It depends on the kind of playing done whether or not this is more or less efficient.

If allocation is a concern (as yoshimi is technically a real-time program, it should be to some extent, though I certainly don't get xruns despite the rampant usage of `new`) then a proper solution would be to avoid allocating notes individually on the heap entirely. I have some ideas for this, including a way to eliminate a few loops in `Misc/Part.cpp` entirely. We'll see if I decide to tackle that; the current implementation doesn't cause any real problems at the moment.

abrolag:
A lot of work here!
Seems fine, and also tested against heavy workloads (not using legato).

src/Params/PADnoteParameters.cpp
```diff
@@ -586,7 +588,9 @@ void PADnoteParameters::applyparameters()
 {
     const int samplesize = (((int)1) << (Pquality.samplesize + 14));
     int spectrumsize = samplesize / 2;
-    float spectrum[spectrumsize];
+    // spectrumsize can be quite large (up to 2MiB) and this is not a hot
+    // function, so allocate this on the heap
+    std::unique_ptr<float[]> spectrum(new float[spectrumsize]);
```
kramlie: +1 on using smart pointers!

PR #87, opened 2020-03-22
Some smaller bug fixes.

kramlie:
I found these while playing around with the new legato implementation.

### DIFF
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -356,9 +356,10 @@ ADnote::ADnote(const ADnote &orig, ADnote *parent, float *parentFMmod_) :
         if (orig.subVoice[i] != NULL)
         {
             subVoice[i] = new ADnote*[orig.unison_size[i]];
+            ADnote *parentVoice = (origVoice != NULL) ? origVoice : this;
             for (int k = 0; k < orig.unison_size[i]; ++k)
             {
-                subVoice[i][k] = new ADnote(*orig.subVoice[i][k], this, freqbasedmod[i] ? tmpmod_unison[k] : parentFMmod);
+                subVoice[i][k] = new ADnote(*orig.subVoice[i][k], parentVoice, freqbasedmod[i] ? tmpmod_unison[k] : parentFMmod);
             }
         }
         else
@@ -367,9 +368,10 @@ ADnote::ADnote(const ADnote &orig, ADnote *parent, float *parentFMmod_) :
         if (orig.subFMVoice[i] != NULL)
         {
             subFMVoice[i] = new ADnote*[orig.unison_size[i]];
+            ADnote *parentVoice = (origVoice != NULL) ? origVoice : this;
             for (int k = 0; k < orig.unison_size[i]; ++k)
             {
-                subFMVoice[i][k] = new ADnote(*orig.subVoice[i][k], this, parentFMmod);
+                subFMVoice[i][k] = new ADnote(*orig.subVoice[i][k], parentVoice, parentFMmod);
             }
         }
         else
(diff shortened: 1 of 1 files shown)


abrolag:
Thanks for picking these up.

Diomendius:
Thanks for catching my mistakes. Clearly, a few things got lost in the boilerplate.

src/Synth/ADnote.cpp
```diff
@@ -220,6 +220,7 @@ ADnote::ADnote(const ADnote &orig, ADnote *parent, float *parentFMmod_) :
             }
         } else {
             vpar.OscilSmp = origVoice->NoteVoicePar[i].OscilSmp;
+            vpar.FMSmp = origVoice->NoteVoicePar[i].FMSmp;
```
Diomendius: I have no idea how I missed this. Thanks.

src/Synth/ADnote.cpp
```diff
@@ -367,9 +368,10 @@ ADnote::ADnote(const ADnote &orig, ADnote *parent, float *parentFMmod_) :
         if (orig.subFMVoice[i] != NULL)
         {
             subFMVoice[i] = new ADnote*[orig.unison_size[i]];
+            ADnote *parentVoice = (origVoice != NULL) ? origVoice : this;
             for (int k = 0; k < orig.unison_size[i]; ++k)
             {
-                subFMVoice[i][k] = new ADnote(*orig.subVoice[i][k], this, parentFMmod);
+                subFMVoice[i][k] = new ADnote(*orig.subVoice[i][k], parentVoice, parentFMmod);
```
Diomendius: Did you check that this actually needed to be fixed? I'm pretty sure I already took this into account.

Near the start of the copy constructor, `origVoice` is set to the parent's `origVoice` if a parent is specified and that parent is not the root:
https://github.com/Yoshimi/yoshimi/blob/56fe5f1db45149acdef61a06a3eebb40957f6892/src/Synth/ADnote.cpp#L155-L156

I probably should've written a comment mentioning this.

Your way might actually be better. I wrote it the way I did because I wanted the copy constructor to only need the immediate parent to be specified, but in hindsight it isn't any worse to have to specify `origVoice` directly, especially since calling the copy constructor on subvoices currently only occurs inside the copy constructor itself.

I suggest moving `parentVoice` to a single assignment before these two if blocks and renaming the `parent` arg to `origVoice_` or similar for consistency, then removing the conditional assignment on lines 155-156 that I quoted above.

In fact, thinking about this makes me wonder if it would cause any problems for `origVoice` to never be null, instead being equal to `this` for the root voice. This should make every `origVoice != NULL` branch unnecessary.

src/Synth/ADnote.cpp
```diff
@@ -714,6 +714,21 @@ void ADnote::legatoFadeIn(float freq_, float velocity_, int portamento_, int mid
...
+        if (!vpar.Enabled)
+            continue;
+
+        if (subVoice[i] != NULL)
+            for (int k = 0; k < unison_size[i]; ++k)
+                subVoice[i][k]->legatoFadeIn(getVoiceBaseFreq(i), velocity_, portamento_, midinote_);
+        else if (subFMVoice[i] != NULL)
+            for (int k = 0; k < unison_size[i]; ++k)
+                subFMVoice[i][k]->legatoFadeIn(getFMVoiceBaseFreq(i), velocity_, portamento_, midinote_);
+    }
```
Diomendius: Good catch. This will also make subvoices fade in and perform portamento as their parent voices do, which is probably what we want.

src/Synth/ADnote.cpp
kramlie: You're right, that's much better. Thanks for cleaning that up! Much of my own confusion in this area stems from the use of the words "orig" to mean two different things. I have launched #90 to address this.

PR #88, opened 2020-04-01
Several more fixes to ADnote legato

Diomendius:
I found a couple of pretty major bugs with my earlier reimplementation of legato (5984640b43c89f83a16242078563721c8d7ba7c7, b2524a649c55dfae6bf89257e9a3f72872e5244c), fixed one bug introduced by Will that only affected legato notes (c3403300b1ec2060fc421860a4ddf1e5d1961886), cleaned up Kristian's fix-that-actually-wasn't (22313e686f59b84b838be6885faeaf616642d8ac), then improved the `origVoice` logic further after a realization I had while looking over Kristian's commit (c1a25a032c6c85afc34a9fc17ef5f009937f53ba).

I also renamed one of the args in one of the ADnote constructors in the header to be consistent with its name in the source file (1cd472df73e652ae904f47a1a537934d28033ce1). It doesn't deserve its own PR, so it ends up here.

### DIFF
--- a/src/Misc/Part.cpp
+++ b/src/Misc/Part.cpp
@@ -451,6 +451,8 @@ void Part::NoteOn(int note, int velocity, bool renote)
         {
             partnote[posb].status = KEY_PLAYING;
             partnote[posb].note = note;
+            partnote[posb].keyATtype = PART::aftertouchType::off;
+            partnote[posb].keyATvalue = 0;
         }
 
         // compute the velocity offset
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -371,7 +371,7 @@ ADnote::ADnote(const ADnote &orig, ADnote *parent, float *parentFMmod_) :
             ADnote *parentVoice = (origVoice != NULL) ? origVoice : this;
             for (int k = 0; k < orig.unison_size[i]; ++k)
             {
-                subFMVoice[i][k] = new ADnote(*orig.subVoice[i][k], parentVoice, parentFMmod);
+                subFMVoice[i][k] = new ADnote(*orig.subFMVoice[i][k], parentVoice, parentFMmod);
             }
         }
         else
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -61,7 +61,7 @@ ADnote::ADnote(ADnoteParameters *adpars_, Controller *ctl_, float freq_,
     forFM(false),
     portamento(portamento_),
     subVoiceNumber(-1),
-    origVoice(NULL),
+    origVoice(this),
     parentFMmod(NULL),
     paramsUpdate(adpars),
     synth(_synth)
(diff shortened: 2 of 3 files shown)


Diomendius:
>I wasn't sure how smart compilers are, so played safe with the magic numbers.

Is this in relation to this comment I made on one of your commits?
https://github.com/Yoshimi/yoshimi/commit/ce00fc1620a79e39d4468652a9ca5dc2cbf87e2a#r38186324
> Compilers perform constant folding, usually even with `-O0` or the equivalent, so there's no need for the `0.168228149f` magic number.
> 
> ```diff
> -        float normalize = synth->oscil_sample_step_f * 0.168228149f;
> +        float normalize = synth->oscil_sample_step_f * (44100.0f / 262144.0f);
> ```
> 
> This should produce identical assembly on any sane compiler. With `-ffast-math` even `synth->oscil_sample_step_f / (262144.0f / 44100.0f)` should produce identical assembly, otherwise for IEEE correctness, a single floating point division opcode is produced instead of a single multiplication one.
> 
> In fact, the `44100.0f` and `262144.0f` magic numbers could be moved somewhere else (declared as `const`/`constexpr`, or `static const`/`static constexpr` if inside a class), and constant folding will still be performed. `262144` is 2^18, and `262144.0f` is `(float)(1 << 18)`, which may be clearer at a glance. Alternatively, `(float)(1024 * 256)` should also work, again producing identical assembly.

I can't say GitHub is that good at organizing conversations. It's great to be able to comment directly on a commit, but such comments easily get lost among the rest of the commits. I guess you got a notification that I had commented on your commit, but somehow got that mixed up with this pull request? I haven't included the change I suggested in that comment in this PR, but I might include it in a future PR if you don't end up changing it yourself.

Anyway, "How smart is the compiler?" seems to come up frequently. While it is usually a bad idea to write code performance-first, sanity-second (["premature optimization is the root of all evil"](https://en.wikipedia.org/wiki/Program_optimization#When_to_optimize)), it always seems to be relevant to have some degree of intuition as to what kinds of problem are "hard to optimize" and what aren't, from the compiler's perspective.

[Constant folding](https://en.wikipedia.org/wiki/Constant_folding) is fairly universal and easy to understand as a rule of thumb: Built-in operations where both operands are known at compile time are all but guaranteed to be performed at compile time, and this applies recursively until one of the operands has a value known only at runtime. The true extent of this is much broader, but that's about as much as I can say in one sentence while being (probably) correct in all cases.

In practice, compilers will do all sorts of algebra and black magic on expressions, but sometimes unintuitive things prevent such optimizations, like floating-point math without `-ffast-math`, where the "correct" result must factor in standards-compliant rounding at each individual step, so `(x * y) * z` is not technically equivalent to `x * (y * z)`, even though they are mathematically equivalent. This relates to constant folding in that without `-ffast-math`, `x * 10.0 * 10.0` generates two separate multiplications while `x * (10.0 * 10.0)` generates one, equivalent to `x * 100.0`. Yoshimi builds with `-ffast-math` by default, so this quirk is not relevant when building Yoshimi with default compiler flags.

abrolag:
Yes. I confess I still struggle with what is advisable and what is best left alone. I'll leave this as it is now - otherwise it could get confusing with two people making changes in the same area. My priority here was getting a unified approach to time related operations instead of the scattered duplication of essentially the same thing.
There are still some buffer size issues I'm not happy about. A number of modifications were made to get LV2 timing behaviour correct. Adjustments were made on top of other adjustments and it became rather a mess - it's still not 100% particularly concerning envelope times. They can be different with different hosts, and more correct with a shorter Yoshimi buffer size than the host one.

PR #90, opened 2020-04-02
Clarify some variables names.

kramlie:
The word "orig" is used to mean two different things, so clean this up
and make it clear:

* "orig" - The original voice when a voice is being cloned into a
  separate, independent instance. Used in legato.

* "parent" - The direct ancestor voice when a voice is being created
  as a result of using sub voices / importing an external voice.

* "top" - The topmost ancestor voice when a voice is being created
  as a result of using sub voices / importing an external voice.

No functionality change.

### DIFF
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -61,7 +61,7 @@ ADnote::ADnote(ADnoteParameters *adpars_, Controller *ctl_, float freq_,
     forFM(false),
     portamento(portamento_),
     subVoiceNumber(-1),
-    origVoice(this),
+    topVoice(this),
     parentFMmod(NULL),
     paramsUpdate(adpars),
     synth(_synth)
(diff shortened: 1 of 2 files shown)


kramlie:
I'd like @Diomendius to double check this before merging!

abrolag:
Understood!
There are still a lot of confusing names, and several different styles. It's difficult to know what (if anything) should be done about it. Apart from near identical names we see things like:
somevariable
Somevariable
someVariable
some_variable
some_Variable
etc. :(

abrolag:
Personally I prefer topVoice - just my 2d (UK old money)

kramlie:
> You should probably change the name in the header, too

Very good point, silly of me to forget! I'll fix it tomorrow, I'm on the wrong computer right now!

kramlie:
Fixed, and all good to go!

abrolag:
Done :)

PR #92, opened 2020-04-05
Undefined behavior fixes

shimpe:
Some C++ problems potentially leading to undefined behavior I found while looking at the code.


abrolag:
One question. We only specify C++11 as a minimum. Is that (rather neat) way of removing duplicates from lists available for that?

shimpe:
Ye. The erase-remove idiom has been a recommended construct for quite some time.

In addition, Yoshimi's CMakeFile seems to enforce C++11 standard, and it compiles and runs with those settings.

https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

abrolag:
Thanks, and thanks for the other fixes. Merging now

PR #93, opened 2020-04-10
Clean up constants, unused headers

Diomendius:
`LegatoMsg` is no longer a thing, so the header which defines it is removed in this PR. Another header actually also defines `LegatoMsg` but was never \#included at all, so it's also removed. (183fd5467c76b798a7437fee04a5527462271b89)

This also moves some magic numbers related to sample rate/oscillator size normalization for modulation into `SynthEngine.h`, and also moves their calculation into `SynthEngine` init. (a5336e881953f66a9f6ce62edfee11d004782f0f, 8a17f0916fe427d39de2c7ac136d78faac93ed3f)

Finally, this improves the way the reference `samplesize` is represented in `PADnoteParameters.cpp`, where it's also used for normalization. `262144.0f` becomes `float(1024 * 256)`. I could've moved it to the header as well, or even reused `oscilsize_ref_f`, but I thought the former would be unnecessary and the latter would be confusing as the numbers aren't related, just equal by coincidence.

This shouldn't affect program behaviour.


abrolag:
Thanks. Had a moment's panic until I remembered to clear the cmake cache!

PR #97, opened 2020-04-18
New algorithm for FM modulation in sub voices with parent modulation.

kramlie:
This one has been bugging me for a while. The old algorithm works well
as long as the frequency and amplitude of the modulator are stable.
However, it has a problem with artifacts when either of those are
changing (especially frequency), such as when using an envelope or an
LFO.

The problem stems from the fact that for each block, the algorithm
always starts at zero phase and works its way to the current phase
using the frequency which is in effect in that block. But this result
can be quite different for the next block if the frequency is
different. The larger the phase offset, the more of this frequency
difference will build up, causing a discontinuity.

The fix is to instead consider the *current* phase to be the center,
and always work backwards and forwards from there. The difference is
easy to hear in a sound with two layers of FM modulation, both with a
frequency or amplitude envelope or LFO.

This results in a rewritten algorithm which:

* Performs better (we start at the right phase instead of having to
  work our way there).

* Requires less memory (the table caching the samples between the zero
  phase and the current phase is no longer necessary).

* Is easier to understand.

* Doesn't break when using envelopes or LFOs on the modulators.

It doesn't get better than that!

### DIFF
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -324,23 +324,31 @@ ADnote::ADnote(const ADnote &orig, ADnote *topVoice_, float *parentFMmod_) :
         {
             if (NoteVoicePar[i].FMEnabled == FREQ_MOD)
             {
-                FMFMoldsmpModded[i] = new float[unison];
-                memcpy(FMFMoldsmpModded[i], orig.FMFMoldsmpModded[i],
+                FMFMoldPhase[i] = new float[unison];
+                memcpy(FMFMoldPhase[i], orig.FMFMoldPhase[i],
                     unison * sizeof(float));
 
-                FMFMoldsmpOrig[i] = new float[unison];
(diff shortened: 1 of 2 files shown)


kramlie:
[Test_Pluck.zip](https://github.com/Yoshimi/yoshimi/files/4496724/Test_Pluck.zip)

You can hear the difference in this sound. Notice the hissing which occurs at various frequencies before the patch is applied.

kramlie:
@abrolag: I know there's a release coming up. This is not urgent at all to merge, so just save it for when you feel it's appropriate. A brand new algorithm may be a bit risky so late in the release process, though I have used it for over a week already with no issues.

abrolag:
I will hold back on this for now. Unfortunately Jesper has hit a bug in the Banks/ FileMgrFuncs code that gives a segfault, suspects it's a string conversion issue :(

Could be a strange directory name on his system, but still is a show stopper!

abrolag:
I can definitely hear the difference. Much cleaner.

kramlie:
Thanks!

PR #100, opened 2020-04-25
Fix constructor related problems

shimpe:
Fixes tons of valgrind warnings while running yoshimi (mostly of the type "jump relies on uninitialized value")


abrolag:
I'm afraid we will have to reject these. This isn't a decision we made lightly and we very much appreciate your interest and efforts.

However, particularly with the case of the constructors, the changes mostly don't actually add anything to the correctness of the code, but change the coding *style* to one that is currently considered 'best practice'. This doesn't take into account that the codebase is nearly 20 years old and has developers who are very familiar with it.

Automated systems have to be used with some caution, as they make no assessment of why a particular action has been taken. Some like the Valgrind suite are very good for finding things like memory leaks, but are not as good as the actual compilers for identifying coding alternatives.

There is a particularly good example of this sort of thing in Interchange.cpp at the start of the mediate() function. At the start of this, 'bool more' is defined but apparently not initialised. However it then *is* initialised at the beginning of a 'do' loop that *must* always be seen at least once before anything can test its value. GCC accepts that as quite OK.

Having said that, there are a few good catches which we will incorporate and acknowledge you as the finder.

shimpe:
Maybe some clarification: Valgrind is only used as detector of problems. It cannot do any refactoring. Clang on the other hand can do refactoring automatically and does so in a way that is guaranteed to be correct by construction - it does not introduce new bugs, only solves bugs.

Initializing everything in the constructor, or at declaration is a known best practice really, avoiding many subtle or very-hard-to-reproduce problems. Maybe the problem isn't there when you write the code, but is silently introduced when changing the code, or reusing it in a new context. I've certainly seen such problems in the current code base, and so did valgrind.

I'm not sure why one would not want to systematically improve the code base - especially if it's 20 years old - but it's your decision to make and I'm not in the mood to argue :)

PR #101, opened 2020-05-10
Don't treat failure to write to GUI thread as an error.

kramlie:
This happens all the time when running as an LV2 plugin, because the
UI is usually not open unless working specifically on the Yoshimi
sound. Most of the time it is just doing its job in the background,
without UI. When this is the case, the events to the GUI are never
consumed and these messages start appearing by the hundreds. This can
be confusing, but it also tends to generate xruns, because printing is
not RT-safe in general. Ignoring these errors is harmless, however,
because the UI will consume them as soon as it is opened.

### DIFF
--- a/src/Interface/InterChange.cpp
+++ b/src/Interface/InterChange.cpp
@@ -1571,10 +1571,7 @@ void InterChange::mediate()
             }
 #ifdef GUI_FLTK
             else if (getData.data.control == MIDILEARN::control::reportActivity)
-            {
-                if (!toGUI->write(getData.bytes))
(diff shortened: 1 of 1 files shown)


kramlie:
@abrolag: What do you think about this one? It's possible to slap a `#ifdef LV2` around it to constrain this to apply only to LV2. However, the whole message doesn't seem terribly useful, because there is no harm if it happens, nor is there typically anything you can do about it if it does happen. The message seems more like a debugging aid for working on UI synchronization.

abrolag:
Yes. One of those ridiculous oversights!

kramlie:
Thanks!

PR #102, opened 2020-05-10
Fix compatibility with lv2 1.18

sebastinas:


src/LV2_Plugin/YoshimiLV2Plugin.cpp
```diff
@@ -379,7 +379,7 @@ bool YoshimiLV2Plugin::init()
 }
 
 
-LV2_Handle	YoshimiLV2Plugin::instantiate (const struct _LV2_Descriptor *desc, double sample_rate, const char *bundle_path, const LV2_Feature *const *features)
+LV2_Handle	YoshimiLV2Plugin::instantiate (const struct LV2_Descriptor *desc, double sample_rate, const char *bundle_path, const LV2_Feature *const *features)
```
kramlie, line 382: Nice cleanup, but you have to lose the `struct` as well, here and the other places where it was changed. On my computer this doesn't compile because `LV2_Descriptor` is a typedef.

sebastinas: Indeed, thanks! Also removed those `struct`s.

abrolag:
Will this still work with older versions of LV2? For example the current debian release carries V 1.14

sebastinas:
Yes, it does.

abrolag:
Thanks for this.
LV2 hasn't had much attention. Indeed we still have a sourceforge feature request going back to 2016. I don't use it myself, and don't think any of the devs do.

kramlie:
I do! Exclusively in fact. :-)

abrolag:
Dammit! me and my big mouth :)

PR #105, opened 2020-07-16
Fix slider value bugs and add new dynamic tooltips

jplloyd:
* Fixes a couple of minor glitches. 
* Refactors the slider color switching code for increased sanity. 
* Adds dynamic tooltips to 5 sliders.
* Switches subsynth harmonic sliders to have the same behavior as the waveform editor harmonic sliders.
That is: you can set/reset them by dragging horizontally while holding down Shift.

* Changes direction of per-harmonic phase slider to be consistent with other vertical sliders, namely:
 up -> increase (positive phase offset)
 down -> decrease (negative phase offset)

The last change should to my knowledge not have any impact on pre-existing midi-learned parameters, other than where the slider handles will end up.

The commit messages provide more details.

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -1,5 +1,13 @@
 yoshimi 1.7.1 M
 
+2020-7-16 Jesper
+* Added dynamic tooltips for AddSynth phase offset sliders,
+  waveform editor per-harmonic phase offset and magnitude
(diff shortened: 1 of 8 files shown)


abrolag:
Thanks for these. Very nice work.

PR #106, opened 2020-07-26
Some fixes I have been saving up

kramlie:
This fixes three bugs:

* Occasional horrible noise when moving phase slider while the voice is playing.
* Unison phase randomness not being respected in sub voices.
* Big optimization when changing LFOs or envelopes with live feedback.

The last one has been particularly important to me. I went from several xruns a minute while changing parameters with MIDI learn, to zero. But it's also a tricky change, which is why I spent a long time testing it before submitting it.


abrolag:
Thanks for these. I appreciate the issues, although I'd not seen them myself, but haven't been playing with the voices much dynamically.

Issue #107, opened 2020-08-03
LV2 plugin require Jack


hfiguiere:
The LV2 plugin require jack, which is kind of a problem.

abrolag:
I wasn't aware LV2 Yoshimi made any such requirements, although I've only ever used it with a Jack backend.
What host are you running, and what OS/engine etc. Version numbers might help too.

hfiguiere:
It requires in the configure checks, and the LV2 includes it. But the problem is that the resulting yoshimi_lv2.so links to libjack, and it's not a good idea.

Here is from `ldd`
```
	libjack.so.0 => /lib64/libjack.so.0 (0x00007f7d718d3000)
```

abrolag:
I still don't understand what you're saying. Are you talking about the build dependencies, or runtime ones?

Currently both ALSA and Jack are required for build, but neither are actually required to run as an LV2 plugin. At least, I can run it in Carla without any input or output - not much use like that of course.

hfiguiere:
`ldd` is what the .so is linked against. So the LV2 plugin requires jack at runtime, as I have shown above. If it doesn't require it, it has no reason to link against `libjack.so.0`

abrolag:
I see what you mean. Removed from the latest commit (build 1852)

hfiguiere:
It still links against `libjack.so.0`

abrolag:
Thanks. I forgot about cmake :(

PR #108, opened 2020-08-08
No libjack for LV2 plugin (#107)

hfiguiere:
This should fix the problem fro #107. I tested it.


abrolag:
Applied manually so closing this with thanks.

PR #109, opened 2020-08-12
Fix detune accumulation bug in SUB synth when altering live voice.

kramlie:
Switch the name `initialfreq` to `basefreq`. This has no functional
impact, but is merely to match the naming in ADsynth; less surprises
for the reader. And then introduce `notefreq` to mean the frequency
which with all detuning applied. ADsynth has no equivalent of
notefreq, it is updated on every cycle. Also rename `setBaseFreq` to
`computeNoteFreq` and remove the argument, to make it more clear that
this function just updates the internal state.

After the renaming, make sure that notefreq is used in all places
where we need the detuned note frequency, and basefreq in all places
where we need the original, unaltered frequency.

### DIFF
--- a/src/Synth/SUBnote.cpp
+++ b/src/Synth/SUBnote.cpp
@@ -58,7 +58,7 @@ const float LOG_0_0001 = logf(0.0001f);
 const float LOG_0_00001 = logf(0.00001f);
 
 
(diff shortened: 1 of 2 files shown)


kramlie:
@abrolag: This fixes all the cases I could find!

abrolag:
Thanks a lot for that. I'll tidy up and get rc1 out :)

PR #111, opened 2020-08-17
Require opts:options feature in turtle manifest

ventosus:


abrolag:
Thank you very much. Greatly appreciated.

PR #119, opened 2020-12-06
Fix last remaining windows that cannot be moved to a second monitor.

kramlie:
plus fix one compile warning.


abrolag:
Thanks for these :)

PR #124, opened 2021-01-06
fixed spelling mistakes

freddii:


abrolag:
While I greatly appreciate all offers of help I must reject this. These are history files, warts and all. We do not try to change history.

PR #128, opened 2021-01-17
Sync LFOs to host or MIDI tempo

kramlie:


kramlie:
@abrolag: Here it is!

As mentioned by email this does not include LFOs for effects. I hope I can get to that later. But all LFOs in the synth engines should be covered.

I had to rebase on top of your UI changes, which produced quite complicated conflicts, so I hope I didn't break anything. I have given the LFO feature a fair bit of testing, but I still think this would benefit from some hardening time in the master branch before we consider a release.

abrolag:
Thanks a lot for this. I agree this should have a thorough test now, to shake out any possible problems. I'll post a note on the user list to see if we can get more people to put it though it's paces.

PR #129, opened 2021-01-18
Make LFO processing work with variable size block buffers (LV2).

kramlie:


abrolag:
Got it. Thanks.

PR #131, opened 2021-01-28
Fix segfault in LV2 instance when displaying scaled MIDI learn window.

kramlie:
It's relying on a count which is living in a different thread. This
causes a race condition when the window is first displayed, where the
count has beeen updated, but the actual elements have not. Therefore,
test each element instead of relying on count.


abrolag:
Thanks for this. Quite unexpected!

PR #132, opened 2021-02-06
Fix live feedback on filters in SUBsynth and PADsynth engines.

kramlie:


abrolag:
That was incredibly quick!

PR #138, opened 2021-05-26
Fix "staircase LFO" bug in LV2 and Jack engine when BPM sync is enabled.

kramlie:
If the internal buffer size is set to be smaller than the buffer size
used by the engine (which is either LV2 processing size or Jack period
size, respectively), then the engine has to call the Synth Engine
several times to fill up its buffer. But this poses a problem if BPM
sync is on, because the beat values are only set before the first
call. This causes a "staircase effect", where the current LFO remains
at or near one beat, until it suddenly jumps to the next beat when the
engine has filled up the buffer and we get to the next full cycle.

The problem is fixed by updating the beat values before each call to
the synth engine, which produces a smooth curve thanks to the
interpolation within each call.

The Alsa engine is not affected by the problem because it forces the
Synth and Alsa buffer sizes to be the same, so there is always exactly
one call to the synth engine.


kramlie:
@abrolag: Just something I discovered while working on the BPM sync for effects.

PR #142, opened 2021-06-27
LFO syncing for effects

kramlie:
The last commit is the main one, with some additional fixes/support commits before that, including a rewrite of the InterpolatedParameter class to support array indexes.

@abrolag: I have not added support for the CLI in this implementation. I can of course do this, but I don't know it so well, and I thought maybe this is a pretty quick thing for you to add? It should be pretty straightforward, just a BPM boolean setting to the effects AlienWah, Chorus, DynamicFilter, Echo and Phaser, as well as a Start setting to all the same effects, except for Echo.

Other than that, I have tested all combinations of Alsa, LV2 and Jack as well as I can, and it seems to be pretty stable for me now.

### DIFF
--- a/src/CLI/CmdInterpreter.cpp
+++ b/src/CLI/CmdInterpreter.cpp
@@ -1376,7 +1376,7 @@ int CmdInterpreter::effects(Parser& input, unsigned char controlType)
             partno = TOPLEVEL::section::insertEffects;
         else
             partno = TOPLEVEL::section::systemEffects;
-        return sendNormal(synth, 0, nFXpreset, controlType, 16, partno,  EFFECT::type::none + nFXtype, nFX);
+        return sendNormal(synth, 0, nFXpreset, controlType, EFFECT::control::preset, partno,  EFFECT::type::none + nFXtype, nFX);
     }
     return REPLY::op_msg;
 }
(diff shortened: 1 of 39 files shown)


kramlie:
Aaaand, I've already discovered the first bug: Pk filter in EQ seems to have lost its effect. Weird, I didn't even change this filter... I need to investigate.

abrolag:
On Sun, 27 Jun 2021 09:53:30 -0700
kramlie ***@***.***> wrote:

>Aaaand, I've already discovered the first bug: Pk filter in EQ seems to have lost its effect. Weird, I didn't even change this filter... I need to investigate.
>
I'm afraid I'm also seeing very strange behavior when running under ALSA
directly. Some effects don't work and some weird thing happens where a track
will suddenly go into overload :(
This doesn't happen in an all jack environment.

Meanwhile I'll take a look at Hermann's cleanup patch.

-- 
Will J Godfrey
http://yoshimi.github.io

kramlie:
> I'm afraid I'm also seeing very strange behavior when running under ALSA directly. Some effects don't work and some weird thing happens where a track will suddenly go into overload :( This doesn't happen in an all jack environment. Meanwhile I'll take a look at Hermann's cleanup patch.

Alright, fixed the filter issue, it was just a small mistake, a copy-paste gone wrong.

However, I'm not sure if it's related to the ALSA issue. It might be, but I would be a bit surprised. Can you still see it if you try this rebased branch? I have not been able to reproduce it on my side.

PR #143, opened 2021-06-27
some further code clean-up for JackEngine

Ichthyostega:
Am 25.06.21 um 21:07 schrieb Will Godfrey:
> In some headers we were still including jack/ringbuffer but this was
> only referenced in the midi struct in jackEngine.h and that struct only
> called for in jackEngine.cpp

Hi Will,

just cross-checked that also with the "references" feature of my IDE.

Not only the `ringBuf` field is unused, also the `pThread` field in the same structure. Thus, the only field in this `midi` strucutre, that is actively initialised and accessed from other code is the `port` field. Moreover, this structure is private in `class JackEngine` and it is never exposed or passed to any other function as such.

Thus we can do a more radical clean-up here: we can remove that structure altogether and just use a plain field of `type jack_port_t`

This has the nice side effect that we can also drop the `#include <pthread.h>`

----

Expanding a little on that, also the field `jack_session_event_t *lastevent`
in the Jack-Session support is unused. And in a similar vein, also the
header `<semaphore.h>` in JackEngine.h and YoshimiLV2Plugin.h is unnecessary.
JackClient doesn't use any semaphores any more, and also YoshimiLV2Plugin
has a private field `_midiSem`, which is no longer referenced any more.

-- Hermann

### DIFF
--- a/src/LV2_Plugin/YoshimiLV2Plugin.h
+++ b/src/LV2_Plugin/YoshimiLV2Plugin.h
@@ -36,7 +36,6 @@
 #include <sys/types.h>
 #include <string>
 #include <vector>
-#include <semaphore.h>
 
 #include "Misc/SynthEngine.h"
 #include "Interface/InterChange.h"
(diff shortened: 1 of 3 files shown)


abrolag:
Thanks for this.
Just tested with standalone and the usual LV2 suspects, with no issues.

-- 
Will J Godfrey
http://yoshimi.github.io

PR #147, opened 2021-08-08
Review Filer Changes: propose to store the FilerLine objects in the STL container

Ichthyostega:
In the review discussion we verified the memory management.
If we use a STL container, it is important to delete the widget objects within this container prior to invoking the FL_Scroll::clear(). FLTK automatically detaches and deletes all widgets within a container widget.

The additions proposed hereby:
 - mark the FilerLine objects as "non-copyable"
 - allocates them directly within the STL container `MasterUI::filerlist`
 - uses a `std::deque<FilerList>`, which is able to deal with non-copyable objects, but still performs reasonably well

### DIFF
--- a/src/UI/MasterUI.fl
+++ b/src/UI/MasterUI.fl
@@ -52,7 +52,7 @@ decl {\#include <string>} {public local
 decl {\#include <iostream>} {public local
 }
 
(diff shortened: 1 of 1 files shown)


abrolag:
Many thanks for this, and the background work.

PR #149, opened 2021-09-25
Fix system effect volume having no effect.

kramlie:
The call to advance the value was missing.

### DIFF
--- a/src/Effects/Alienwah.cpp
+++ b/src/Effects/Alienwah.cpp
@@ -68,6 +68,8 @@ Alienwah::~Alienwah()
 // Apply the effect
 void Alienwah::out(float *smpsl, float *smpsr)
 {
+    outvolume.advanceValue(synth->sent_buffersize);
+
     for (int i = 0; i < synth->sent_buffersize; ++i)
     {
             smpsl[i] += float(1e-20); // anti-denormal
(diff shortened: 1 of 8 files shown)


kramlie:
@abrolag: Here you go!

Btw, can you confirm that the Pan knob in System effects is wired opposite of what it should be? Left is right, and right is left, at least in the distortion effect.

abrolag:
Seems to be right for Reverb but takes a *very* long time to change sides :)
Not availalbe for EQ.
OK for the rest - depending on settings.

abrolag:
On Sat, 25 Sep 2021 07:26:40 -0700
kramlie ***@***.***> wrote:

***@***.***: Here you go!
>
>Btw, can you confirm that the Pan knob in System effects is wired opposite of what it should be? Left is right, and right is left, at least in the distortion effect.


Thanks for getting on to this so quickly.

Pan seems OK here, but it is *very* slow changing sides on reverb.

-- 
Will J Godfrey
http://yoshimi.github.io

kramlie:
> Pan seems OK here, but it is *very* slow changing sides on reverb.

Oh, now it's ok. What happened..? I must have crossed the wires in Jack or something. All good!

PR #150, opened 2021-10-02
LV2 fix: Mixer Audio Send dropdown disabled when it shouldn't be.

kramlie:
It depends on the value of the preferred audio driver, but this is
irrelevant when running with LV2. Instead we need to check for whether
the LV2 instance is Multi or Single.


abrolag:
Interesting! Never thought of that.

PR #153, opened 2021-11-22
"power"-Branch

Ichthyostega:
Several refactorings and small adjustments as part of the Yoshimi-Testsuite effort.
**Note**: this changeset might cause minute changes to the computed numbers; the intention is to rearrange computations in a way to be more stable on different compilers / optimisers.

Generally speaking, the goal of this effort is to establish a Suite of automated tests to detect unintended changes affecting the generated sound or the runtime of Synth computations.

_See also the Repository_ https://github.com/Ichthyostega/yoshimi-test


abrolag:
Thanks for these. I've a few more midilearn cleanups to do, then I'll set rc11

PR #161, opened 2022-03-10
Implement fallback BPM.

kramlie:
```
commit 597ee541e63aaeac72e1c404c8fbba53ac99e029
Author: Kristian Amlie <kristian@amlie.name>
Date:   Mon Mar 7 23:44:15 2022 +0100

    Implement fallback BPM.
    
    It is available from the master window, and provides a way to set the
    global BPM when the MIDI driver does not provide it, for example
    because there is no ALSA device available to send MIDI clock
    messages. It becomes active whenever this information is missing.
    
    For both Jack and LV2, it is trivial to add support for this. For ALSA
    it's more complicated, because we need to detect that a certain window
    has passed before we can conclude that MIDI clock information is
    missing.
    
    As part of this, I rewrote the ALSA MIDI loop to be more efficient by
    avoiding constant looping and polling. Instead, use the system call
    `poll()`, where we can both sleep and set a timeout at the same
    time. This has the advantage of reacting instantly to events, but also
    provides a long time where the CPU can sleep if there are no
    events. I've set the timeout to 500ms so that quitting will still be
    relatively quick.
    
commit 0ad4538c4dea793dcdab4a7a2a5c2c959ccefe86
Author: Kristian Amlie <kristian@amlie.name>
Date:   Mon Mar 7 19:04:36 2022 +0100

    Fix BPM timing issues when audio/midi drivers are mixed.
    
    `getBeatValues()` should not be used as a basis for calculating new
    beat values, since they try to calculate the beat values according to
    the current time, for use by the audio thread. In those cases we
    should rather get the raw values, without timing measurements.
    
```

### DIFF
--- a/src/Interface/Data2Text.cpp
+++ b/src/Interface/Data2Text.cpp
@@ -1131,6 +1131,9 @@ string DataText::resolveMain(CommandBlock *getData, bool addValue)
         case MAIN::control::keyShift:
             contstr = "Key Shift";
             break;
+        case MAIN::control::bpmFallback:
+            contstr = "Fallback BPM";
+            break;
         case MAIN::control::mono:
             contstr = "Master Mono/Stereo ";
             showValue = false;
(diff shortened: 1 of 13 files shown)


abrolag:
Thanks a lot for this. As I've not touched the Undo branch for a few days I'll merge this straight into master

PR #162, opened 2022-04-21
Padthread - wavetable background build

Ichthyostega:
The "padthread" feature allows to rebuild the Wavetable for PADSynth as background task.

- can continue to play while new wavetables are prepared
- can automatically trigger background build on relevant parameter changes
- smooth crossfade transition to the new wavetable on all playing notes
- PADsynth can re-trigger itself periodically, with a random walk

Implementation required extensive rework of Yoshimi's internals, esp. the memory management for FFT data, since generation of libfftw3 computation plans is not threadsafe.
All spectrum and waveform data is now managed by data holder classes with unique ownership; data is arranged in memory in the right order for Fourier Transform, and thus
no new allocation and copy is necessary.


abrolag:
Excellent work!

PR #166, opened 2022-05-25
Fix risc64 build as reported by Bo YU in debian

snd1:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1011451


abrolag:
Thanks for this.

PR #169, opened 2022-08-20
Fix missing phase randomness when importing sub voices.

kramlie:
This was working for normal oscillators at one point, but was broken
when the live editing was introduced. For modulators I believe it has
never worked. Depending on how the sound is set up, the effect can be
very subtle, so it took me a long time to discover it.

However, there is one scenario where it is very audible. Set the
second voice to import from the first voice (doesn't matter which
waveform), and set unison of the second voice to 50. Make sure phase
randomness is at 100%. This produces a loud volume spike at the
beginning of the sound because all the waveforms' phases line up
exactly, before they start drifting apart. This should not happen when
randomness is on.

This required several fixes. First and foremost, we were not even
storing the phase offset for modulators, so introduce that in the
constructor.

Second, a problem is that the assignment of phases in the parent voice
was happening too late; the sub voice was created, and its phase
assigned, before the parent voice had calculated any phase. Because of
live editing, most of the phase calculations now happens in
computeNoteParameters(), which executes on every cycle. However, it
would be unreasonably expensive to call this early, because we still
have to call it again as part of the normal cycle. Then we would have
to call it twice for every note, and it's not a cheap function.

So instead, split out the part that deals with the phases into
separate functions, and call them when we need them instead. This is
still two calls, but much cheaper than computeNoteParameters().

And finally, the code was restructured slightly to be make sure we
call synth->numRandom() only once, while the rest of the calculation
we can call as many times as we want. Otherwise every cycle would have
a new random phase, which is just noise.

### DIFF
--- a/src/Synth/ADnote.cpp
+++ b/src/Synth/ADnote.cpp
@@ -148,11 +148,13 @@ ADnote::ADnote(ADnoteParameters& adpars_, Controller& ctl_, Note note_, bool por
     , topVoice{topVoice_}
     , parentFMmod{parentFMmod_}
 {
+    // Start phase for sub-Voices should be negative so that the zero phase in
+    // the first cycle will result in a positive phase change.
     int phase = (topVoice==this)? 0 : synth.oscilsize - phaseOffset;
-             // Start phase for sub-Voices should be negative
-             // so that the zero phase in the first cycle will result in a positive phase change.
-    for (int nvoice = 0; nvoice < NUM_VOICES; ++nvoice)
+    for (int nvoice = 0; nvoice < NUM_VOICES; ++nvoice) {
         NoteVoicePar[nvoice].phaseOffset = phase;
+        NoteVoicePar[nvoice].fmPhaseOffset = phase;
+    }
 
     construct();
 }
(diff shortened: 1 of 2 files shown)


kramlie:
I tested it a fair bit when I first made it, and ran with it for several weeks. But I did the work prior to 1f15996308c02c868610a853c9a5492680ea8a5f, and therefore I had to solve some pretty hefty conflicts afterwards. I hope I didn't make any mistakes..

abrolag:
On Sat, 20 Aug 2022 08:28:14 -0700
kramlie ***@***.***> wrote:

>I tested it a fair bit when I first made it, and ran with it for several weeks. But I did the work prior to 1f15996308c02c868610a853c9a5492680ea8a5f, and therefore I had to solve some pretty hefty conflicts afterwards. I hope I didn't make any mistakes..
>
Can you detail your exact setup? I tried applying this to a side branch, but
can't detect the spike you referred to with either this or the current master.

-- 
http://yoshimi.github.io

abrolag:
On Sat, 20 Aug 2022 21:52:44 +0100
Will Godfrey ***@***.***> wrote:

>On Sat, 20 Aug 2022 08:28:14 -0700
>kramlie ***@***.***> wrote:
>
>>I tested it a fair bit when I first made it, and ran with it for several weeks. But I did the work prior to 1f15996308c02c868610a853c9a5492680ea8a5f, and therefore I had to solve some pretty hefty conflicts afterwards. I hope I didn't make any mistakes..
>>  
>Can you detail your exact setup? I tried applying this to a side branch, but
>can't detect the spike you referred to with either this or the current master.
>
Just want to add I'm not rejecting the pull request - just can't match the
symptoms.

-- 
It wasn't me! (Well actually, it probably was)

... the hard part is not dodging what life throws at you,
but trying to catch the good bits.

kramlie:
Uhm.. So honestly I'm quite confused now. I can't reproduce it either, using my own instructions! Not even using my original branch point.

What is clear though, is that it does produce a spike when the import happens on the modulator side. Use the same instructions, but enable Morph modulation in the second voice, import the first voice in the modulator, and set both Vol. and V.sns. to maximum.

I must have simply mixed up the instrument, and maybe Modulation was still enabled when I thought it was off. Still, I don't really understand how I missed it...

As for the patch, I suggest that we merge it as is anyway though. It does fix the Modulator case, and since I moved both calculations to their own functions, I think it's a good cleanup, even though it doesn't technically fix anything for normal oscillators.

abrolag:
OK, I'm happy with that.

kramlie:
Thanks!

PR #170, opened 2022-08-29
Remove unneeded idle thread from LV2 plugin.

kramlie:
The thread does nothing but spin and use CPU time. This makes an
enormous difference on big projects where you have many Yoshimi
instances.

### DIFF
--- a/src/LV2_Plugin/YoshimiLV2Plugin.cpp
+++ b/src/LV2_Plugin/YoshimiLV2Plugin.cpp
@@ -281,16 +281,6 @@ void YoshimiLV2Plugin::processMidiMessage(const uint8_t * msg)
 }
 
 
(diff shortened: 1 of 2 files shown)


kramlie:
Hey @abrolag, this should be very safe, I've been running it for almost six months already, I just kind of forgot I had this patch in my tree! 🙂

abrolag:
On Mon, 29 Aug 2022 11:57:52 -0700
kramlie ***@***.***> wrote:

>The thread does nothing but spin and use CPU time. This makes an
>enormous difference on big projects where you have many Yoshimi
>instances.
>
>Signed-off-by: Kristian Amlie ***@***.***&gt;
Thanks for that. I'd been puzzled by it for some time (hence the comment!)

-- 
http://yoshimi.github.io

PR #176, opened 2023-02-21
Typos

luzpaz:
Fixes various typos in the codebase.


abrolag:
I'm accepting this pull, although there was an error when you'd missed out an important part of a line in the midi-learn file. There is also one correction where you miss the point of a joke so I'll revert that :)
"Shirley Knot" as a reference the a joke in the 'Airplane' comedy film!

PR #177, opened 2023-04-06
Performance improvements in sortResultsThread.

kramlie:
```
commit bd64fca3c46c8e17c238ca6da0de075dc3af5a06
Author: Kristian Amlie <kristian@amlie.name>
Date:   Thu Apr 6 14:09:13 2023 +0200

    solo: Replace custom timekeeping with `clock_gettime`.
    
    This should improve both performance and accuracy since the kernel is
    in charge of the ticks.
    
commit 958f5237234d98f82d63a40da74247c276e2d353 (mine/sortResultsThread)
Author: Kristian Amlie <kristian@amlie.name>
Date:   Thu Apr 6 14:10:58 2023 +0200

    Spin sortResultsThread only when we need to, instead of all the time.
    
    This improves performance significantly, especially when using
    multiple instances where most are idle.
    
```

### DIFF
--- a/src/Interface/InterChange.cpp
+++ b/src/Interface/InterChange.cpp
@@ -117,6 +117,8 @@ InterChange::InterChange(SynthEngine *_synth) :
     undoStart = false;
     cameFrom = envControl::input;
     undoMarker.data.part = TOPLEVEL::section::undoMark;
+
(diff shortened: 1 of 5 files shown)


abrolag:
Thanks for the Kristian - nice bit of tidying up.

kramlie:
My pleasure!

PR #178, opened 2023-04-23
legato: Fix incorrect allocation size in cloneArray, leading to crash.

kramlie:
It worked fine as long as the unison size was eight or less, since that's the number of voices, but not for 9 and above.


abrolag:
A sneaky one! Thanks for thefix.

PR #179, opened 2023-04-25
Fix old LV2 state data not being saved correctly after loading.

kramlie:
The problem is that the Pname is set to DEFAULT_NAME, which then prevents saving later, even though the sound is correct at the time of the load. Most likely we should never set it to DEFAULT_NAME as long as it isn't explicitly set to that in the XML.

This is a very destructive bug as it is not apparent that it has happened until the next time you load.


abrolag:
Thanks for this. I notice I made a comment that it is all rather complicated!

kramlie:
And you are not wrong! 😀

PR #180, opened 2023-04-25
Fix incorrect header in state data when saving with LV2.

kramlie:
I don't think it has any effect currently, but doesn't hurt to do it right.


abrolag:
I don't think this bit of code has changed since Andrew first created it!

PR #181, opened 2023-05-09
Review: minor improvements to evelope handling

Ichthyostega:
Envelope points were switched to float values with changeset... 9db835a32

Did a code review and verified with the Testsuite...
Existing presets/instruments will be migrated to float values without affecting sound.

Proposing some minor code tweaks, and especially also store the free-form envelopes as float values


abrolag:
Thanks for this. Very nice work.

PR #194, opened 2023-10-11
Fix inability to restore windows to secondary screens.

kramlie:
We need to take multiple screens into account when calculating width,
height and position, otherwise they will always be restored to the
center screen.

Also fix checkSane being called too late on Adsynth Voice window.

### DIFF
--- a/src/UI/ADnoteUI.fl
+++ b/src/UI/ADnoteUI.fl
@@ -2826,8 +2826,8 @@ o->hide();}
             fetchW = voiceDW;
             fetchH = voiceDH;
         }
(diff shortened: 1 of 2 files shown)


kramlie:
@abrolag: Just a small fix for a bug that's been annoying me for a while!

abrolag (approved):
Thanks for these

abrolag:
On Wed, 11 Oct 2023 10:47:09 -0700
kramlie ***@***.***> wrote:

>@abrolag: Just a small fix for a bug that's been annoying me for a while!
>

both done, thanks :)

-- 
http://yoshimi.github.io

PR #197, opened 2024-02-15
Fix incorrect phase in ADnote Voice parameters window.

kramlie:
This broke in 7cc3efd02dc3be73b because of a small detail: When reading from the voiceOscillatorPhase control, we can't use the raw value. We need to subtract it from 64, which is how the original Poscilphase gets set. Without this fix the oscillator display is always one half cycle off when compared to the oscillator window.


abrolag:
Good call - and I thought I was being careful with these :(

PR #200, opened 2024-03-06
Gui-connect-1

Ichthyostega:
<h1>Transfer core-data into the GUI</h1>

The goal of this development effort is to remove direct data access by the GUI.
Up to now it is a common pattern for the GUI to reach into internal data of the Core,
which both causes problems with thread safety and hinders a separation of GUI and Core.
While commands and response messages are already dispatched through a command system
based on ringbuffers, there was no suitable way to transfer larger data blocks.

- This feature-branch introduces a connection scheme with push-updates through a new component **GuiDataExchange**. 
- start-up of the UI is changed to be attached to the `InterChange`, and using an **InterfaceAnchor** provided through the new GuiDataExchange scheme
- as first application, all **Effect-UIs** changed from direct data access to push-updates for all parameters and for the response-curve of the EQ

This feature is reasonably complete, but leaves some further clean-up to be done.
Moreover, there are some known bugs, which were identified during this reworking...
- the EQ response curve shape is not generated consistently by the EQ effect
- on all but the first Effect-component, spurious differences to the preset are marked on the widgets, while in fact there is no difference to the preset


abrolag:
Thanks for this work.

PR #203, opened 2024-06-30
Find manual: don't descend into any network mounts

joostn:
In case there is a network filesystem mounted somewhere under /home/, don't descend into this when finding the manual.

On my computer it starts searching an SMB mount containing many gigabytes of files just to find the manual.

From 'man find':
-xdev Don't descend directories on other filesystems.


abrolag:
Thanks for this - good catch

PR #207, opened 2024-09-16
Instance Manager

Ichthyostega:
This is part II of an effort to rework and improve the communication between GUI and core.

Part I was merged as pull #200 this spring, and introduced new core infrastructure, the GuiDataExchange. Moreover, Effect parameters and EQ graphs were already switched over to the new communication scheme. However, integrating this new scheme based on **push-updates** with the existing code to handle SyntEngine instances turned out to be near impossible. Thus, the top-level of the Yoshimi application and LV2 plugin was rewritten to introduce an **Instance Manager**.

This sweeping change reorders any start-up and shut-down activities and moves all code dealing with the creation of `SynthEngine`, `Config` instances and the IO-back-ends (`MusicClient`, `MusicIO`) into the new `InstanceManager` singleton, which can now be accessed as `Config::instances()`. This allowed to introduce a well defined **Lifecycle**, and the _event processing_ in the »main thread« now processes a _duty cycle_ to check and advance lifecycle state of all SynthEngine instances.

The start-up of the FLTK GUI could be re-arranged to fit into this new scheme. We do not need to rely on FLTK-messaging any more, rather, when reaching the end of the `BOOTING` phase, the `MasterUI` is created from the »main thread« and a new _bootstrap message_ is sent to the Synth-Engine thread to complete start-up. At this point, an initial slate of push-updates is generated and sent up to the UI to prime the UI-connection.

Moreover, as part of these changes, a lot of code-modernisation was carried out, since Yoshimi has recently switched up to the C++17 standard. We now use the »chrono« framework and the »mutex« wrappers from the standard library, and -- in an ongoing effort -- a lot of direct usages of `new` and `delete` could be replaced by _smart-pointers._


src/Misc/Util.h
```diff
@@ -0,0 +1,116 @@
...
+
+#include <string>
+
+namespace std {// forward declarations to avoid pervasive includes
+
+  template<typename T>
+  class allocator;
+  template<typename K, typename CMP, class ALLO>
+  class set;
+
```
nekopsykose, line 32: this doesn't build against libc++:  

```
/usr/include/c++/v1/set:1486:43: error: reference to 'set' is ambiguous
 1486 | using set _LIBCPP_AVAILABILITY_PMR = std::set<_KeyT, _CompareT, polymorphic_allocator<_KeyT>>;
      |                                           ^
/usr/include/c++/v1/set:564:28: note: candidate found by name lookup is 'std::__1::set'
  564 | class _LIBCPP_TEMPLATE_VIS set {
      |                            ^
/builddir/yoshimi-2.3.3/src/Misc/Util.h:31:9: note: candidate found by name lookup is 'std::set'
   31 |   class set;
      |         ^
```

just include `<algorithm>` and `<set>` like normal...

Ichthyostega: Thanks for reporting.
Which version of the library / compiler are you using?

Btw, mindlessly including some header in the whole code base can significantly increase built times, and thus turnaround. OTOH, STL headers are typically well modularised and not that problematic.

Do you really also need `<algorithm>` to make it work in your setup?

nekopsykose: > Which version of the library / compiler are you using?

it's llvm libc++ 18 (and the compiler is clang 18, but only the stl matters for that)

> Do you really also need <algorithm> to make it work in your setup?

it's the correct header for `std::find`. it might work without it but that would be relying on a (transitive) include to define it, which is prone to breaking in the future when the stl headers get shuffled around and less things are automatically included, as libc++ happens to like doing every other release (ironically, so less stuff is automatically included, and compile times would improve where those things are not really needed)

Ichthyostega: As pointed out in the related [issue #108](https://github.com/Yoshimi/yoshimi/issues/208), I made this optimisation _prematurely_ (since in similar situations I had seen quite significant gains). But in this case an actual measurement (with `#include <set>` everywhere) revealed no tangible impact, presumably because we are including several more heavyweight STL headers in many compilation units downstream.

In the recent Bugfix release v.2.3.3.1 we have now removed this problematic forward declaration and just use the includes in the regular way. Sorry for the breakage

nekopsykose: no worries :) thanks for the changes and your work on yoshimi!

PR #211, opened 2024-11-08
Avoid uninitialized error by making sure `interchange` is constructed.

kramlie:
When initializing `rootCon`, it uses `interchange`, but it's not constructed until a few lines later. Initialization is always done in member order, so we need to rearrange the members to fix it.


Ichthyostega:
Hi Kristian,
good catch! This inconsistency probably slipped in during the investigation how to bring up all parts in various config variants, which also involved a lot of shuffling in the way how the base config is pulled up.

First, the situation _is not actually broken_, since the initialisation of the root-connection only uses a static registration function (I was careful to do it this way in order to avoid too much tangling in constructors). Yet needless to say, the fact that the initialisation invokes a pass-through function on the not-yet initialised interchange is a lurking danger, and should be fixed!

One caveat though: In your proposed Fix, you also moved the Runtime (=Config object instance) down. This makes me feel queasy, given how deeply interwoven the Config object is with everything else in the system.

My proposal would be to keep the uniqueID and the Runtime as first subobjects to be initialised, then do it in the order you proposed, and place the rootCon object indeed into a  new `private:` member section as you propose.

src/Misc/SynthEngine.h
```diff
@@ -69,18 +69,20 @@ using std::unique_ptr;
 
 class SynthEngine
 {
-        const uint uniqueId;
-        Config Runtime;
-
-        GuiDataExchange::Connection<InterfaceAnchor> rootCon;
```
Ichthyostega, line 75: ...so my proposal would be _only_ to move rootCon down into a new `private:` section

abrolag:
On Fri, 08 Nov 2024 09:27:53 -0800
Hermann Voßeler ***@***.***> wrote:

>@Ichthyostega commented on this pull request.
>
>
>
>> @@ -69,18 +69,20 @@ using std::unique_ptr;  
> 
> class SynthEngine
> {
>-        const uint uniqueId;
>-        Config Runtime;
>-
>-        GuiDataExchange::Connection<InterfaceAnchor> rootCon;
>
>...so my proposal would be _only_ to move rootCon down into a new `private:` section
>

I just pushed a minor update to master before seeing this. It's only
documentation so shouldn't make any difference.

I had been concerned about seeing non-initialisation warnings from GCC so
presumably those were the same issue.


Hmmm. All the emails seem to be overlapping! I'll go with whatever you both
agree with :)

-- 
http://yoshimi.github.io

kramlie:
@Ichthyostega: Thanks for looking! How about now?

Ichthyostega:
looks all fine for me; the danger of using a yet uninitialised `interChange` is addressed.

abrolag:
Thanks for this

PR #212, opened 2024-11-25
Prevent unlimited unison count on layered sub voices.

kramlie:
This should have been there since the introduction of sub voices, I just forgot to add a ceiling check back then.

Fun fact: Before this fix, if you try to set all 8 voices to have 50 unison count, and stack them as sub voices on top of each other, it tries to allocate nearly forty trillion unison buffers, and several hundred billion voices. It might be tough to run that realtime!


abrolag:
Yes, this could be 'interesting!

PR #214, opened 2024-12-14
Increase LV2 compatibility by supporting "UI" base class methods.

kramlie:
The interfaces "showInterface" and "idleInterface" are meant as fallbacks if none of the sub class methods work or are supported. This enables the jalv standalone host to launch Yoshimi, which wasn't possible before.

Note that support for the sub class extension "external-ui" is still preferred, since without it Yoshimi has no way to tell the host that the UI was closed. So we keep that as the main class with the "UI" base class as a fallback.

### DIFF
--- a/src/LV2_Plugin/YoshimiLV2Plugin.cpp
+++ b/src/LV2_Plugin/YoshimiLV2Plugin.cpp
@@ -668,7 +668,7 @@ YoshimiLV2PluginUI::~YoshimiLV2PluginUI()
 
 bool YoshimiLV2PluginUI::init()
 {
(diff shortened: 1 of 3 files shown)


kramlie:
I've been reading up on LV2 lately! 🙂

abrolag:
Thanks :)

abrolag:
On Sat, 14 Dec 2024 04:25:26 -0800
kramlie ***@***.***> wrote:

>I've been reading up on LV2 lately! 🙂
>

Steady on! Don't peak too soon :)

-- 

PR #216, opened 2025-01-30
Several fixes to LV2 programs

kramlie:


kramlie:
@abrolag: Hey, this fixes the corruption issue in the LV2 program list. Does it help on the Muse issue you were having?

kramlie:
The PR also improves on the LV2 programs list; it now respects your config settings better, and it can also use programs in different roots now. See the second commit for details.

abrolag:
Very nice work. Thanks for tracking this down.

abrolag:
On Thu, 30 Jan 2025 05:50:12 -0800
kramlie ***@***.***> wrote:

>@abrolag: Hey, this fixes the corruption issue in the LV2 program list. Does it help on the Muse issue you were having?
>
Well done on this sleuthing!

Unfortunately it doesn't fix the problem I'm seeing with Muse.
However I'll try to rebuild the test file again from scratch, in case there is
something weird about the original file.

Would you like be to send a compressed copy of the original file, so you can
have a look at it yourself?

Compressed it's 59k

-- 
Will J Godfrey.

PR #218, opened 2025-02-03
LV2 fixes

kramlie:
This fixes several issues with LV2 that were discovered after more work and testing following my last PR. See individual commits for details on each fix.


abrolag:
Thanks for this. I see where you're going. However it doesn't resolve the Muse problem.

PR #226, opened 2025-06-01
MXML(1) : Rework XML handling

Ichthyostega:
The breaking API and ABI changes introduced with MXML v4.x pose a challenge, since Yoshimi must be able to support both versions of this widely used base library for several years to come. This was taken as an opportunity to loosen the direct connection between Yoshimi core logic and MXML data structures — creating an internal interface that can be implemented differently for each library variant.

To support the task of disentangling Yoshimi core logic and XML handling, the existing implementation in `class XMLwrapper` was taken apart and most of the internal state tracking was removed, since lib-MXML actually provides navigation operations and a ref-count based memory management.

A new interface was created for the internal code of Yoshimi to work with XML.
While the old code was based on a _stateful navigation_ through the XML tree,
the new interface follows the »DOM style«
- `XMLStore` is now the access point, manages metadata and can be created
   + as empty tree marked with an »XML-type«
   + by loading from a file
   + by parsing the contents of a char buffer
- The actual data access is now provided by `XMLtree`, which is a lightweight smart-ptr and refers to some XML-subtree; further nested elements will be represented by further `XMLtree` instances


abrolag:
Thank you for all this work.

PR #229, opened 2025-07-09
Fix channel aftertouch using the wrong parameter as basis.

kramlie:
This results in several bugs, both setting the wrong parameter during the aftertouch, and restoring it to the wrong value afterwards.


kramlie:
Yeah. Probably modifying the Q with aftertouch is far less common than the cutoff.

PR #235, opened 2025-08-26
improve metainfo

umlaeute:
`appstreamcli validate desktop/metainfo/yoshimi.metainfo.xml` showed a couple of problems, which I am fixing with this PR:
- use https://yoshimi.sourceforge.io as the base-URL for all documentation/homepage (rather than a mixture of sourceforge.io, sourceforge.net and github.io; also use https:// throughout)
- fix links to URLs (some filesystems are case-sensitive)
- fix the HTML in the `description` tag

like always, I've tried to separate all the changes into multiple commits, so if you do not agree with a specific change, just cherry-pick whatever you want :-)


abrolag:
Thanks for this - it's easy to forget metadata :(

PR #241, opened 2025-12-05
Fix start phase bug for frequency LFOs.

kramlie:
A bit surprising that this bug hasn't been reported before. It requires setting a start phase which happens to end up in the lower quarter of the phase cycle. In this case the frequency would be less than zero, which the rest of the code is not prepared to handle. This would result in the negative magnitude of the response being too strong (meaning note would be too low pitched), before suddenly jumping back once it is corrected by the phase wrapping code. Easier to hear with a slow LFO.


abrolag:
Thanks for this. it's surprising this one has stayed hidden!

PR #248, opened 2026-03-11
Fix empty bank file after failed startup

Ichthyostega:
This problem seems to have been caused by an interplay of various recent changes

- changes to the start-up sequence done two years ago (for the `GuiDataExchange`)
- changes to the configuration and start-up / shutdown to reliably persist the banks
- changes in preparation for MXML 4 : the new XMLStore is always initialised with headers and can thus be persisted even without adding data.

The first problem was stale state in MusicClient, after unsuccessfully probing an IO backend. This caused the "null backend" erroneously to fail. The second problem was that the bank file could be saved before the SynthEngine was fully initialised, whenever "Yoshimi staged a strategic retreat"

### DIFF
--- a/src/Misc/Bank.h
+++ b/src/Misc/Bank.h
@@ -175,6 +175,7 @@ class Bank
         size_t generateSingleRoot(string const& newRoot, bool clear = true);
 
         uint getVersion() {return version; }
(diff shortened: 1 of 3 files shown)


abrolag:
Thanks for fixing these rather sneaky bugs!

